PR #6, opened 2016-04-26
Re-define the API of the library

philpl:


src/map.js
```diff
@@ -79,55 +76,96 @@ export default class ReactMapboxGl extends Component {
...
+
+    const didCenterUpdate = (
+      this.props.center !== nextProps.center &&
+      nextProps.center.reduce((acc, x, key) => {
+        if (!center[key] || center[key] !== x) {
+          return true;
+        }
+
+        return acc;
+      }, false)
```
alex3165: Is it not what `.equals` is doing ?

src/map.js
```diff
@@ -46,24 +46,21 @@ export default class ReactMapboxGl extends Component {
   getChildContext = () => {
     return {
       map: this.state.map
-    } 
+    }
   };
 
   state = {};
 
-  _onStyleLoaded(map) {
```
alex3165: Thanks i forgot to remove this

philpl:
@alex3165 :shipit:

PR #15, opened 2016-05-13
Add support for MultiPolygon geojson objects

nbadag:
Was working with a geojson file of US state outlines and discovered a number of shapes weren't rendering properly because they were of `type: "MultiPolygon"` (as opposed to `"Polygon"` used in the 'All Shapes' example). Mapbox GL appears to support MultiPolygons just fine, so seemed like a pretty straightforward fix.


alex3165:
Thanks for your PR 👍

PR #38, opened 2016-08-13
Adding pitch default value

ideoma:


ideoma:
Map went nuts in example London map with Mapbox gl 0.21.0 until I tracked the problem to pitch default.

alex3165:
thanks ! 👍

PR #44, opened 2016-09-07
Add "raw" GeoJSON layer

philpl:


philpl:
@alex3165 In case someone wants to render raw GeoJSON, which is something that I need for adding this lib to nteract

alex3165:
cool 👍

PR #47, opened 2016-09-27
Add missing mapbox GL removal for cleaning up

### LINKED ISSUE #46: Mobile browsers crashing after a few map loads
I don't know if this has to do with garbage collection or running out of Web GL contexts...it seems if you load and kill 3 map instances, your page will crash mobile Safari. iOS 10.

philpl:
Fix #46


philpl:
@alex3165 Should be fine. Can be included in the next release, I guess

PR #49, opened 2016-10-03
Add support for AttributionControl position

joshwcomeau:
Hi!

In a personal project, I needed to position the mapbox attribution in the top-right corner, instead of the bottom-right. So I added it as a prop. 

No worries if you're not interested in accepting this PR, but I thought I'd open it in case others could benefit.

The implementation was pretty straightforward. The only bit I'm unsure of is whether it's best to have an `attributionPosition` prop (as I've done), or to mirror the official API more closely with an `attributionControl` prop that takes an object.

I've opted for simplicity over accuracy, but let me know if it should be flipped :)

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -32,6 +32,11 @@ import ReactMapboxGl from "react-mapbox-gl";
   - `jumpTo`
   - `easeTo`
   - `flyTo`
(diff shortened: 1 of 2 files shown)


alex3165:
Thank you for the PR I will test it asap 👍

PR #51, opened 2016-10-06
Update mapbox-gl to 0.25.1

nrako:
Hi @alex3165,

Have you considered updating mapbox-gl to the latest version, lots of new improvements and features, would you consider this PR? If not, did I missed an issue?

Thanks

### DIFF
--- a/example/webpack.config.js
+++ b/example/webpack.config.js
@@ -8,7 +8,7 @@ module.exports = {
   ],
   devtool: 'cheap-module-eval-source-map',
   resolve: {
-    extensions: ['', '.js', '.jsx'],
+    extensions: ['.js', '.jsx'],
     alias: {
       'react-mapbox-gl': path.join(__dirname, '../lib/index.js')
     }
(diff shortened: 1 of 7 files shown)


nrako:
Don't accept this, this doesn't work. @alex3165 do you have any idea how to tackle this? Since mapbox-gl now recommend to use the distribution build and removed "support" for webpack I'm not sure how this should be solved. Maybe this package should only require "mapbox-gl" as a `peerDepenendencies` and let people alias `mapbox-gl` to the distribution build in their bundler config. WDYT?

alex3165:
Hello @nrako Thank you for your help, it is funny because we recently solved the problem with Webpack by building the library ourselves but indeed now we have to do a step back and simply use their bundle and directly do something like ` import MapboxGl from "mapbox-gl"; ` instead of using `../vendor/mapbox-gl.bundle`. I think mapbox-gl should be a `dependency` and not a `peerDependencies` as it already is. If you want to tackle this it would be very appreciated, feel free to ask any question in this issue feed https://github.com/alex3165/react-mapbox-gl/issues/53

Issue #57, opened 2016-10-30
Why zoom property is in array? Disabling Moving Method?


Altiano:
1. I think zoom property should be just primitive float, I don't see any reason to put it in array as it also different from Mapbox GL API

2. How to disable zoom method? I don't want to use any moving method.

alex3165:
Yes I know it is an array on purpose, it was a Float before but in order to check for a reference equality on the map component it is required to pass it through an array : 
```
    const didZoomUpdate = (
      this.props.zoom !== nextProps.zoom &&
      nextProps.zoom !== map.getZoom()
    );
```

In this case if `this.props.zoom` is `[4]` and `nextProps.zoom` is a new array with a Float 4 inside it will break the equality because it does a reference check but it won't in case it is an Float.

If you have a good idea of how to solve it nicely I am happy to discuss of it.

Altiano:
Maybe I'm not really understand what you're trying to say but as far as I know, if you're just want to comparing it, just use `==` (i.e. don't use strict equality `===`)

alex3165:
Even with a `!=` it doesn't break the reference equality as if I pass an array because Int, String and Float are primitive types.

philpl:
@Altiano We chose to change this back in version 0.8. It was a quick and easy fix to make sure that the zoom style was updated reliably, every time it's actually changed. Otherwise when the zoom level stays constant the styles are not updated.

The point is, we can't tell whether the zoom has changed, when the number itself has the same value. We can work around this in an easy way like this, by wrapping the number in an array, which is allocated  anew each time when you create one (note: `[] !== []`). This is not exactly fixing the problem, but we didn't see a reason to fix it properly and waste time on it.

To elaborate on equality in JS, `===` & `!==` stand for reference equality. It is technically not a strict equality, but to simplify it, it checks whether the values that are being pointed to are the same instance, so technically "in the same point in memory" (not being highly precise, since this doesn't always hold up).

I'm getting a bit off track here, but to further continue `==` & `!=` are type inferring before checking value equality. Details here: https://dorey.github.io/JavaScript-Equality-Table/

cc @alex3165

DvStephen:
This is pulling up some history I appreciate, but I can't follow the reasoning here since you've essentially made a constant expression by strictly comparing an array this will always be true that they do not match, regardless of the values within them.

Is there an original context for this change?

Wykks:
Everything is explained in the comment above by kitten; and still makes sense.

PR #63, opened 2016-11-04
Compare primitive data types for camera options

mtauer:
When I was using the component with redux together the zoom array and the center object where recreated although the camera option values haven't changed.

My solution was to compare the primitive data types instead of the composite data types inside the `componentWillReceiveProps` method.

Also the comparisons `nextProps.zoom !== map.getZoom()` and `nextProps.center !== map.getCenter()` were comparing differently structured data types.

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -217,19 +217,19 @@ export default class ReactMapboxGl extends Component {
     const bearing = map.getBearing();
 
     const didZoomUpdate = (
-      this.props.zoom !== nextProps.zoom &&
-      nextProps.zoom !== map.getZoom()
+      this.props.zoom[0] !== nextProps.zoom[0] &&
+      nextProps.zoom[0] !== map.getZoom()
     );
 
     const didCenterUpdate = (
-      this.props.center !== nextProps.center &&
-      nextProps.center !== map.getCenter()
+      (this.props.center[0] !== nextProps.center[0] || this.props.center[1] !== nextProps.center[1]) &&
+      (nextProps.center[0] !== map.getCenter().lng || nextProps.center[1] !== map.getCenter().lat)
     );
 
     const didBearingUpdate = (
       this.props.bearing !== nextProps.bearing &&
       nextProps.bearing !== map.getBearing()
-    )
+    );
 
     if (didZoomUpdate || didCenterUpdate || didBearingUpdate) {
       map[this.props.movingMethod]({


alex3165:
Hey @mtauer, thanks for your PR but the zoom equality is done with an array on purpose in order to do a reference equality. Please refer to https://github.com/alex3165/react-mapbox-gl/issues/57 for more informations.

mtauer:
Hi @alex3165, I was not aware of the issue of the zoom values comparison and, to be honest, I also don't know exacly what zoom style updates mean that are mentioned in [#57](https://github.com/alex3165/react-mapbox-gl/issues/57). But as far as I understand the code there are two problematic comparisons still left in the code:

1) `nextProps.zoom !== map.getZoom()` - https://github.com/alex3165/react-mapbox-gl/blob/master/src/map.js#L221
This comparison will always return true as `nextProps.zoom` is an array and `map.getZoom()` is a float.

2) `nextProps.center !== map.getCenter()` - https://github.com/alex3165/react-mapbox-gl/blob/master/src/map.js#L226
This comparison will always return true as well as `nextProps.center` is an array of two floats and `map.getCenter()` is an object of a `lat` and a `lng` value.

3) Additionally I'm really interested in knowing more about the the issue when the camera options need to be be updated although the zoom (float value) hasn't changed. Would be cool if you could quickly point me to an example.

alex3165:
1) I have tested it you are completely right we need to fix it :
```
    const didZoomUpdate = (
      this.props.zoom !== nextProps.zoom &&
      nextProps.zoom[0] !== map.getZoom()
    );
```

2) Same you are right it is a mistake :
```
    const didCenterUpdate = (
      this.props.center !== nextProps.center &&
      nextProps.center !== Object.values(map.getCenter())
    );
```

3) For example when the zoom of the map changed `getZoom() // return 3` and you want your props to take over and update the zoom value even if `this.props.zoom === nextProps.zoom // 4 for example` then in this case you just have to break the reference equality and it will change the zoom value of the map. Then this way the zoom value will be `4` if it does make sense. You can try on the London cycling example.

On the London cycle example you can try to click on a station and change the zoom with your mouse and click again on a station, you expect the zoom to focus on the station but the value you passed through the props didn't change with the previous one so you just have to break the reference equality to update the zoom value of the map.

This can be achieved with an object as well and maybe it make more sense than an array ...


Do you want to open a PR for the 2 first points ?

mtauer:
Thanks for the examples ...will test them.

I just created a [new pull request](https://github.com/alex3165/react-mapbox-gl/pull/64). The zoom comparison I did exactly the same, for the center comparison, I would compare the properties individually as Object.values creates a new array and also returns the values in arbitrary order.

alex3165:
@mtauer Thank you I have merged it 👍

PR #74, opened 2016-11-29
Implement Popup and Marker using React in order to keep track of the context

alex3165:


src/marker.js
```diff
@@ -15,52 +20,57 @@ export default class Marker extends React.Component {
...
 
   render() {
-    const onClick = this.props.onClick || (() => {});
-    const style = {
-      transform: overlayTransform(this.state),
-      zIndex: 2,
-      cursor: (this.props.onClick ? 'pointer' : 'auto'),
+    const { onClick, style } = this.props;
+
+    const finalStyle = {
```
kmamykin: I think what got lost with this change is the logic to assign cursor: 'pointer' only if there was this.props.onClick passed. Looks like now style needs to be explicitly passed in, which is cool. However the default may need to be cursor: 'auto' since onClick is not required, and without optional properties the marker should not appear to be clickable?

src/popup.js
```diff
@@ -5,61 +5,77 @@ import {
...
 
   componentDidMount() {
     const { map } = this.context;
+
     map.on('move', this.handleMapMove);
-    map.on('moveend', this.handleMapMoveEnd);
+    map.on('moveend', this.handleMapMove);
     // Now this.container is rendered and the size of container is known.
     // Recalculate the anchor/position
     this.setState(overlayState(this.props, this.context, this.container));
```
kmamykin: in Marker component this like is changed to `this.handleMapMove()`

kmamykin:
Checking in to see if there is anything else I can do to help get this merged in.

alex3165:
@kmamykin Yes I just want to make sure there is not warning in the london cycle example when using the new Popup component, Once it is done as it is a React component without dependencies to `mapbox-gl-js` I would like to write some unit tests for the Popup and the Marker, I have introduced Jest recently to the project.

alex3165:
@philpl and @kmamykin I think it is ready, can you guys review it ? :)

PR #86, opened 2016-12-22
Add events to the marker and the popup

alex3165:


alex3165:
@philpl Can you review this 👍

example/server.js
```diff
@@ -3,13 +3,15 @@ var webpack = require('webpack')
 var config = require('./webpack.config')
 
 var compiler = webpack(config)
+var port = 8080;
```
philpl: Can you replace these two lines with `process.env.X || DEFAULT`? Just for convenience

example/server.js
```diff
@@ -3,13 +3,15 @@ var webpack = require('webpack')
...
 
 var server = new WebpackDevServer(compiler, {
   publicPath: config.output.publicPath,
   historyApiFallback: true,
   noInfo: true
 })
 
-server.listen(8080, "localhost", function() {
-  console.log('☕️  Server is listening on localhost:8080.');
+server.listen(port, host, function() {
```
philpl: By default `host` shouldn't probably be passed at all

src/marker.js
```diff
@@ -12,12 +12,11 @@ export default class Popup extends React.Component {
     anchor: OverlayPropTypes.anchor,
     offset: OverlayPropTypes.offset,
     children: PropTypes.node,
```
philpl: It'd be great to get all the missing callback `propTypes` in? Is there any reason why they were missing in the first place?

PR #87, opened 2016-12-22
Re-compute coordinates on componentWillUpdate

alex3165:


alex3165:
@philpl Can you review this please

src/projected-layer.js
```diff
@@ -51,6 +51,17 @@ export default class ProjectedLayer extends React.Component {
...
   }
 
+  componentWillUpdate(nextProps) {
+    const { coordinates } = this.props;
+
+    if (
+      coordinates[0] !== nextProps.coordinates[0]
+      || coordinates[1] !== nextProps.coordinates[1]
+    ) {
+      this.setState(overlayState(nextProps, this.context.map, this.container))
```
philpl: You cannot call `setState` inside `componentWillUpdate`. This should use `componentWillReceiveProps`.

alex3165: Indeed, good catch, I have just fixed it

PR #90, opened 2017-01-01
Fix Popup className and default anchor props

### LINKED ISSUE #89: Popup anchor does not setup automatically when anchor is null
By default anchor should setup automatically.
To make this work one way is to do this:

in popup.js
```
 render() {
...
    const anchorPrefix = 'mapboxgl-popup-anchor-';

    if (anchor) {
      defaultClassName.push(`${anchorPrefix}${anchor}`);
    }

    return (
      <ProjectedLayer
       ...
        anchorPrefix={anchorPrefix}>
        <div className="mapboxgl-popup-tip"></div>
        <div className="mapboxgl-popup-content">
          { children }
        </div>
      </ProjectedLayer>
    );
```

and in projected-layer.js
```
  render() {
    const {
     ...
      className,
      anchorPrefix,
      ...
    } = this.props;
    const { anchor } = this.state;

    ...
    const finalClassName = `${className} ${anchorPrefix}${anchor}`;

    return (
      <div
        className={finalClassName}
       ...>
        { children }
      </div>
    );
  }
```
Cheers

philpl:
Fix #89

### DIFF
--- a/src/popup.js
+++ b/src/popup.js
@@ -5,8 +5,6 @@ import {
   OverlayPropTypes,
 } from './util/overlays';
 
-const defaultClassName = ['mapboxgl-popup'];
-
 export default class Popup extends React.Component {
   static propTypes = {
     coordinates: PropTypes.arrayOf(PropTypes.number).isRequired,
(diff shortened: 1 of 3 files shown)


alex3165:
Nice 👍

PR #96, opened 2017-01-10
fix name of marker class

mollymerp:
just noticed this change as a result of https://github.com/alex3165/react-mapbox-gl/commit/4f7a4ddac7a34f38fc78e25ba5c4d2cdbd3fdbe1 and I think you meant to keep the name `Marker`. 

thanks!

### DIFF
--- a/src/marker.js
+++ b/src/marker.js
@@ -6,7 +6,7 @@ import {
 
 const propsToRemove = { children: undefined };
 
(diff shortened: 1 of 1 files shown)


alex3165:
It didn't technically break anything as it is a default export but obviously it was wrong. Thanks for the PR and happy to see someone from Mapbox having an interest on this library.

PR #103, opened 2017-01-17
raster layer enabled and source component

z0d14c:
This PR makes it so that the following is valid:

```
<Source id="example_id" sourceOptions={RASTER_SOURCE_OPTIONS} />
<Layer type="raster" id="example_id" sourceId="example_id" />`
```

Will add documentation before merging but wanted to put this up for public scrutiny.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ The library include the following elements :
 - ZoomControl
 - ScaleControl
 - Popup
+- Source
 
 ## How to start
 
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ The library include the following elements :
 
 - ReactMapboxGl
 - Layer
+- Source
 - GeoJSONLayer
 - Marker (Html marker)
 - Feature
@@ -17,7 +18,6 @@ The library include the following elements :
 - ZoomControl
 - ScaleControl
 - Popup
-- Source
 
 ## How to start
 
--- a/src/source.js
+++ b/src/source.js
@@ -2,10 +2,6 @@ import React, { PropTypes } from 'react';
 
 export default class Source extends React.Component {
 
-  render() {
-    return null;
-  };
-
   static contextTypes = {
     map: PropTypes.object,
   };
@@ -34,4 +30,9 @@ export default class Source extends React.Component {
       map.removeSource(this.id);
     }
   }
+
+  render() {
+    return null;
+  };
+
 }
--- a/src/source.js
+++ b/src/source.js
@@ -1,0 +1,37 @@
+import React, { PropTypes } from 'react';
+
+export default class Source extends React.Component {
+
+  render() {
+    return null;
+  };
+
+  static contextTypes = {
+    map: PropTypes.object,
+  };
+
+  static propTypes = {
+    id: PropTypes.string,
+    sourceOptions: PropTypes.object,
+  };
+
+  id = this.props.id;
+
+  source = {
+    ...this.props.sourceOptions,
+  }
+
+  componentWillMount() {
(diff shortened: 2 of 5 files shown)


z0d14c:
added documentation.

alex3165:
Thanks for the PR, let me know when you are ready for review

I am wondering if we shouldn't move all the source logic from the `Layer` to the `Source` component but it would involve a complete redesign of the `Layer` component. I am talking about https://github.com/alex3165/react-mapbox-gl/blob/master/src/layer.js#L208.

Although, having a `Source` component to only add a Mapbox source could be a first step toward this goal.

z0d14c:
@alex3165 i think im ready for review when you are. its pretty basic but i would be in favor of an initially basic functionality while we mull over the prospect of a complete redesign of Layer.

src/source.js
```diff
@@ -0,0 +1,37 @@
...
+    const { map } = this.context;
+    if (!map.getSource(this.id)) {
+      map.addSource(this.id, this.source);
+    }
+  }
+
+  componentWillUnmount() {
+    const { map } = this.context;
+    if (map.getSource(this.id)) {
+      map.removeSOurce(this.id);
```
philpl: A small typo here ;)

src/source.js
```diff
@@ -0,0 +1,37 @@
...
+    return null;
+  };
+
+  static contextTypes = {
+    map: PropTypes.object,
+  };
+
+  static propTypes = {
+    id: PropTypes.string,
+    sourceOptions: PropTypes.object,
```
philpl: I like that this mirrors the `Layer` props 👍

src/source.js
```diff
@@ -0,0 +1,37 @@
...
+  render() {
+    return null;
+  };
+
+  static contextTypes = {
+    map: PropTypes.object,
+  };
+
+  static propTypes = {
+    id: PropTypes.string,
```
philpl: We have to have this as a required prop

philpl:
@alex3165 Sources are really simple and we already had the logic in place to safely "eject" and use a custom `sourceId`. We might need to disallow adding children to the Layer component if the user has passed a `sourceId` prop.

@z0d14c Reviewed this really quickly

z0d14c:
@philpl i think ive addressed your requested changes! good catch on the typo

README.md
```diff
@@ -17,6 +17,7 @@ The library include the following elements :
 - ZoomControl
 - ScaleControl
 - Popup
+- Source
```
alex3165: Move `Source` underneath the `Layer`

src/source.js
```diff
@@ -0,0 +1,37 @@
+import React, { PropTypes } from 'react';
+
+export default class Source extends React.Component {
+
+  render() {
```
alex3165: Move render to the bottom to keep it consistent.

philpl: @z0d14c I'm kinda interested why you've moved it to the top actually, since the common best practice is to put it at the bottom. :)

Also, Are you the Zodiac Killer?

z0d14c: I'm a little bit new to React (coming from a background of mainly AngularJS 1.x and node) so I'm still learning some of the best practices (this is a great example).

I am not, although I do murder code sometimes 👿

tsemerad:
Thanks for doing this PR. I've been using `react-map-gl` for awhile, and just today started trying out this library to compare. I quickly hit this issue of setting multiple Layers for a single dynamic source. I think this should do the trick.

z0d14c:
@alex3165 @philpl i think i addressed all of your requested changes now!

PR #107, opened 2017-01-20
Fix pitch prop not being processed in componentWillReceiveProps

tsemerad:
…enter, zoom, and bearing updates.


philpl:
LGTM :ship: Thanks for the fix 😄

dorthwein:
Looks like this is still an issue after the first update for bearing.  First change does adjust bearing, subsequent changes does not.

alex3165:
Hey @dorthwein if you find any issue related to a weird behaviour on the update of the pitch / bearing / zoom or coordinates feel free to open an issue 👍

PR #113, opened 2017-02-01
Add maxDecimalPlaces to Marker component

VilliGunn:
There is a blur issue with Chrome and Safari when the translate property has a pixel value with decimal numbers.

I added a prop for the Marker component that controls how many, if any decimal places should be used. Not sure the control option is needed, maybe the maxDecimalPlaces should just be set to `0` by default.

Here I am not using maxDecimalPlaces:
blur

Here I am using maxDecimalPlaces:
noblur

You will need to click on the images to get the full res image.

Both images from Chrome 55.0.2883.95

Please feel free to make changes or comments on this :)

### DIFF
--- a/example/src/cluster.js
+++ b/example/src/cluster.js
@@ -59,6 +59,7 @@ export default class ClusterExample extends Component {
                 key={key}
                 style={styles.marker}
                 coordinates={feature.geometry.coordinates}
+                maxDecimalPlaces={0}
                 onClick={this.onMarkerClick.bind(this, feature.geometry.coordinates)}>
                 M
               </Marker>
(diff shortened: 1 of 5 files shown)


alex3165:
Thanks for the PR, is there a considerable position shift if you max zoom to a `Marker` and check with `toFixed(0)` and without ?

VilliGunn:
Here you can see a `Marker` with and without decimal numbers on max zoom

With decimal:
decimal

Without decimal:
nodecimal

I can't see any considerable position shift for this specific `Marker` at least. If there are no decimals the `Marker` can only move max 0.5 pixels on both the X and Y axis.

alex3165:
Ok, then I guess when can go with a `toFixed(0)` by default 👍  would you mind doing the change ?

alex3165:
Awesome thank you 👍

PR #115, opened 2017-02-05
Source component should respond to sourceOptions.data updates

ericrwolfe:
Right now if you create a `<Source>` component with GeoJSON data, the data can't be updated after the component is mounted.

This would allow you to update the GeoJSON data e.g. `<Source id="source-id" sourceOptions={{type: 'geojson', data: geojsonData}}>` and have mapbox-gl re-render when the data changes.

### DIFF
--- a/src/source.js
+++ b/src/source.js
@@ -31,6 +31,22 @@ export default class Source extends React.Component {
     }
   }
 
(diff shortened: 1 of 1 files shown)


alex3165:
Look good, thanks for the update

PR #119, opened 2017-02-09
update source to compare tile urls

z0d14c:
For projects that modify the source urls, it is necessary for the source to be added and removed with the updated source urls (there is no mapbox-gl-js function to simply changeUrl, unfortunately).

### DIFF
--- a/src/source.js
+++ b/src/source.js
@@ -13,6 +13,21 @@ export default class Source extends React.Component {
 
   id = this.props.id;
 
+  compareTileUrls(nextTileUrls) {
+    const tileUrls = this.props.sourceOptions.tiles;
+    let changed = false;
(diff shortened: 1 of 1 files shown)


alex3165:
Hey @z0d14c sorry for the late reply I was quite busy, do you mind rebasing from master ? thanks

philpl:
This is gonna conflict with the ts migration, I think? :laughing: maybe we should take this over?

z0d14c:
Please feel free to take this over if its convenient for you to convert to
typescript

alex3165:
@z0d14c Can you please try with this : https://github.com/alex3165/react-mapbox-gl/pull/122

I will close this PR then we can continue to speak about this feature on this feed https://github.com/alex3165/react-mapbox-gl/pull/122

PR #122, opened 2017-02-13
Update source on componentWillReceiveProps

alex3165:

PR #126, opened 2017-02-17
Bug/before prop watch

z0d14c:
Note: this feature branched from #122 so it has some changes that are showing in the diff from that branch.

The interesting changes (its like 3 lines) are in `layer.ts`

Summary: We need to compare prop.before so that it gets updated in `componentWillReceiveProps`

### DIFF
--- a/src/layer.ts
+++ b/src/layer.ts
@@ -193,7 +193,7 @@ export default class Layer extends React.PureComponent<Props, void> {
   }
 
   public componentWillReceiveProps(props: Props) {
-    const { paint, layout } = this.props;
+    const { paint, layout, before, id } = this.props;
     const { map } = this.context;
 
     if (!isEqual(props.paint, paint)) {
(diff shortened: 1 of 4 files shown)


alex3165:
can you rebase from master ? so that the diff is correct please ?

src/layer.ts
```diff
@@ -211,6 +211,10 @@ export default class Layer extends React.PureComponent<Props, void> {
         map.setLayoutProperty(this.id, key, layoutDiff[key]);
       });
     }
+
+    if (props.before !== before) {
+      map.moveLayer(id as string, props.before);
```
alex3165: You should use `this.id` and not `this.props.id`

alex3165: Also no need for casting id to a string

alex3165:
I have pushed the changes on master if you don't mind I will close this PR 👍  thanks for the help. https://github.com/alex3165/react-mapbox-gl/commit/2277b22573dbd59147affa9fa64d50357f8e5113

Issue #144, opened 2017-03-03
GeoJSONLayer - Update Layout and Paint


daniel5151:
`<Layout>` will check if layout / paint have changed between updates, while `<GeoJSONLayer>` does not. 
I haven't looked into it too much, but i'm fairly certain that all that needs to be done is to add:
```js
    const { paint, layout,  before } = this.props;
    const { map } = this.context;

    if (!isEqual(props.paint, paint)) {
      const paintDiff = diff(paint, props.paint);

      Object.keys(paintDiff).forEach((key) => {
        map.setPaintProperty(this.id, key, paintDiff[key]);
      });
    }

    if (!isEqual(props.layout, layout)) {
      const layoutDiff = diff(layout, props.layout);

      Object.keys(layoutDiff).forEach((key) => {
        map.setLayoutProperty(this.id, key, layoutDiff[key]);
      });
    }
```

to `<GeoJSONLayer>`'s `shouldComponentUpdate`

daniel5151:
After looking into it a bit more, I realize that my proposed solution does not actually work 😅.
I've got it working for fillPaint, but i'm fairly sure extending it to the 7 other props shouldn't be difficult at all.

```js
    const { id } = this;
    const { data, fillPaint } = this.props;
    const { map } = this.context;

    if (props.data !== data) {
      (map.getSource(id) as MapboxGL.GeoJSONSource).setData(props.data);
    }

    // import `isEqual` and `diff`
    if (!isEqual(props.fillPaint, fillPaint)) {
      const fillPaintDiff = diff(fillPaint, props.fillPaint);

      Object.keys(fillPaintDiff).forEach((key) => {
        map.setPaintProperty(this.id + "-fill", key, fillPaintDiff[key]);
      });
    }

    // ... and so on and so forth
```

PR #148, opened 2017-03-09
GeoJSONLayer Enhancements

daniel5151:
I found the GeoJSONLayer component to be a bit under-featured, so I enhanced it a bit:

- Added support for mapbox's fill-extrusion layer (for displaying 3D buildings and such)
- Updated `componentWillReceiveProps` to update the map on any changes to the `[layer]Paint` and `[layer]Layout` props
- Changed it so that by default, GeoJSONLayer will not render any layers that do not have explicitly defined Paint / Layout props

This resolves issue #144 

(Please ignore that last PR, I committed with the wrong account 😅. This is my first PR, cut me some slack 😄)


package.json
```diff
@@ -62,6 +62,7 @@
     "react-dom": "^15.0.1"
   },
   "devDependencies": {
+    "@types/core-js": "^0.9.36",
```
alex3165: I don't think this is needed

src/geojson-layer.ts
```diff
@@ -1,25 +1,39 @@
 import * as React from 'react';
 import * as MapboxGL from 'mapbox-gl/dist/mapbox-gl';
+const isEqual = require('deep-equal');
+import diff from './util/diff';
 import { generateID } from './util/uid';
 import { Sources, SourceOptionData } from './util/types';
 
+const type_to_layer_LUT = {
```
alex3165: Can you use camel case ?

package.json
daniel5151: `npm run build` would fail for me until I added this dep.

alex3165:
Awesome 👍  do you think you could write some test for geojson-layer as well ? if not it is fine we can first merge this and add some tests later

daniel5151:
Tests would be good, but I don't really have the time to do them now. Sorry 😞

alex3165:
Also can you remove the yarn.lock ? don't think this should go out on this PR, it can be added separately later

package.json
alex3165: Ok fair enough 👍

alex3165:
I have checked the GeoJson example on your branch and there is an issue the points doesn't appear on the map anymore.

daniel5151:
Ah, yes, the example... I forgot about that.
Off the top of my head, the reason they are not showing up is because I changed the default behaviour of the component to not render any layers that have not been given Paint/Layout props.

I'll update the example to show off those hidden layers too.

docs/API.md
```diff
@@ -140,8 +140,8 @@ import { GeoJSONLayer } from "react-mapbox-gl";
 
 ### Properties
 - **data** *(required)* : `String | Object` The url to the geojson file or the geojson file itself.
-- **lineLayout** | **symbolLayout** | **circleLayout** | **fillLayout** : `Object` Layer layout informations. [mapbox layout api](https://www.mapbox.com/mapbox-gl-style-spec/#layer-layout)
-- **linePaint** | **symbolPaint** | **circlePaint** | **fillPaint** : `Object` Paint informations. [mapbox paint api](https://www.mapbox.com/mapbox-gl-style-spec/#layer-paint)
+- **lineLayout** | **symbolLayout** | **circleLayout** | **fillLayout** | **fillExtrusionLayout** : `Object` Layer layout informations. [mapbox layout api](https://www.mapbox.com/mapbox-gl-style-spec/#layer-layout)
```
philpl: (Unrelated to this line)

can you remove the example `yarn.lock` file? I know it's unnecessary and we all love yarn, but unless the entire project supports it, we might not want to create any confusion ;)

src/geojson-layer.ts
```diff
@@ -1,25 +1,39 @@
 import * as React from 'react';
 import * as MapboxGL from 'mapbox-gl/dist/mapbox-gl';
+const isEqual = require('deep-equal'); //tslint:disable-line
```
philpl: @alex3165 Are there typings for deep equal? if not, can we add some?

src/geojson-layer.ts
```diff
@@ -50,8 +64,11 @@ export default class GeoJSONLayer extends React.Component<Props, void> {
     const layerId = `${id}-${type}`;
     layerIds.push(layerId);
 
-    const paint: Paints = this.props[`${type}Paint`] || {};
-    const layout: Layouts = this.props[`${type}Layout`] || {};
+    const paint: Paints = this.props[`${typeToLayerLUT[type]}Paint` ] || {};
+
+    // deafult undefined layers to invisible
```
philpl: small typo

docs/API.md
alex3165: He already removed the yarn.lock file

src/geojson-layer.ts
alex3165: no typings for deep-equal indeed, we could write some or use `equal.lodash` instead maybe

docs/API.md
alex3165: Actually it is here again, @daniel5151 please add it to the `.gitignore`

daniel5151:
That's weird... why is Travis failing? I don't think this latest commit should have affected anything.

alex3165:
@daniel5151 Something is wrong in the latest node typings file:
```
node_modules/@types/node/index.d.ts(50,14): error TS2687: All declarations of 'iterator' must have identical modifiers.
```

Can you try an older version of typings for node ?

daniel5151:
That seems to have done the trick

src/geojson-layer.ts
```diff
@@ -50,8 +64,11 @@ export default class GeoJSONLayer extends React.Component<Props, void> {
     const layerId = `${id}-${type}`;
     layerIds.push(layerId);
 
-    const paint: Paints = this.props[`${type}Paint`] || {};
-    const layout: Layouts = this.props[`${type}Layout`] || {};
+    const paint: Paints = this.props[`${typeToLayerLUT[type]}Paint` ] || {};
+
+    // default undefined layers to invisible
+    const visibility = Object.keys(paint).length ? 'visible' : 'none';
```
alex3165: Hey, having a second look at the PR and I am wondering, why do you default to invisible ? Can you convince me it is an expected behaviour by default ?

alex3165: Ok I get it, can you please explain the new behaviour in the description of the `GEOJson` component on the documentation ?

philpl:
@alex3165 I'd say this is ready to go for now?

alex3165:
Thanks @daniel5151 nice job and thanks @philpl for the review 👍

PR #152, opened 2017-03-14
call setData when children disappear

jluxenberg:
Previously, if you had a Layer that had Features as children,
and then the props changed such that the Layer no longer had
children, <source>.setData would not be called to remove the
Features from the source.

To fix this, I've changed Layer so that it will always call
setData when rendered. This should be fine because Layer#render
will only be called if the component should update.

I've added tests as well.


alex3165:
Thanks @jluxenberg nice job 👍

PR #157, opened 2017-03-18
GeoJSONLayer accepts layerOptions

jluxenberg:
layerOptions was missing from GeoJSONLayer, this adds it. I've also added a test.

Tested manually as well, I think this is a very safe change.

### DIFF
--- a/src/__tests__/geojson-layer.test.tsx
+++ b/src/__tests__/geojson-layer.test.tsx
@@ -1,0 +1,60 @@
+import * as React from 'react';
+import GeoJSONLayer from '../geojson-layer';
+import { withContext } from 'recompose';
+import { mount } from 'enzyme';
+
+describe('GeoJSONLayer', () => {
+  let GeoJSONLayerWithContext: any;
+  let addLayerMock = jest.fn();
+
+  beforeEach(() => {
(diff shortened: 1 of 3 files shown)


src/__tests__/geojson-layer.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+      minzoom: 13,
+      paint: {'fill-color': '#001122'}
+    }, undefined]);
+  });
+
+  it('Should call addLayer when no layerOptions provided', () => {
+    mount(
+      <GeoJSONLayerWithContext
+        fillPaint={{'fill-color': '#001122'}}
+        data={{ "type": "FeatureCollection", "features": []}}
```
alex3165: Can you define a const for `fillPaint` and `data` and use it for both tests ? Also the linter doesn't check the test I will fix it but can you use single quote 👍

alex3165:
@jluxenberg Thanks for the PR nice job with the test ! Can you add `layerOptions` to the documentation as well and do the small changes required above ? Then it should be good to go.

jluxenberg:
@alex3165 Thanks for taking a look! I've made the changes you requested.

PR #158, opened 2017-03-20
Add className props

sohibul:
It would be more flexible to style if I can pass className to the container. For example I can style the two button as children using className.


alex3165:
Thanks for the PR, can you add it to the documentation as well please ?

sohibul:
Done, I'm sorry I'm newbie in contributing open source

alex3165:
@sohibul No problem, can you please add `className` prop to the documentation on this PR please

PR #168, opened 2017-03-28
Bug fix in Source; Update deps and examples;

lamuertepeluda:
Hi there,

sorry for the cumulative PR.

I began this because I started having trouble with source while feeding it with geojson data.  
I was clustering geojson data like in [this mapbox example](https://www.mapbox.com/mapbox-gl-js/example/cluster/) rather than with `<Cluster>`. This approach was working fine before v1.0.0.  
Digging into this issue, I found out that the data are not loaded when a new Source is added.
I therefore added a listener on the map 'load' event into source, together with another property `onSourceLoaded` which perhaps make more sense than `onSourceAdded`.

I also found out that this repository currently includes a dist, minified version of mapbox (which was also out of date), that makes debugging quite hard when it's MapBox fault, and also runtime type checking impossible (e.g. `typeof x === LatLng`will always fail because the returned type is minified).  
It took me a while to figure it out (also because MapBox doesn't officially support webpack and uses Flow instead of TS), but eventually I found out how to include mapbox sources into the project, thanks also to [folks at Uber](https://github.com/uber/deck.gl/commit/c7115415ec978c9d5c47910435c2339cc5a15a78).

So here is what you get with this PR:

1. compile mapbox from source (v0.34.0 [which also fixes several bugs on their side](https://github.com/mapbox/mapbox-gl-js/blob/master/CHANGELOG.md#0340-march-17-2017))
1. add mapbox typings
1. update overall deps
1. add a new `ClusterWithLayers` example which follows the mapbox clustering approach (looks more efficient to me)
1. Fix bug in the `all-shapes` example where `setState` was being called also after the component was unmounted.

Hope I didn't forget anything. Thank you for your great work. I will try to help you with other issues as well.


lamuertepeluda:
I also forked https://github.com/lamuertepeluda/react-mapbox-gl-debug in order to use my fork, even if there is one manual build step to be added since I didn't release the package.

alex3165:
Hey @lamuertepeluda Thank you for the PR, I quickly had a look and noticed you are using typings but we don't want to use typings to manage the type definitions, I am using npm for that :). I will have a better look at the PR later this week

lamuertepeluda:
@alex3165 sorry I'm still far from being a TypeScript black belt 😅 , but it was only trying to get it working and compiling in vs code with mapbox sources instead of minified dist file.

I should have fixed it now.

lamuertepeluda:
@alex3165 Sorry to bug you, but did you have a chance of looking at this PR?

alex3165:
@lamuertepeluda The PR contain lot of changes I need to test, I will have a closer look this week end. Sorry for the delay, I want to make sure everything make sense especially on the dependencies you added.

alex3165:
Hey @lamuertepeluda, I had a closer look and it would be better to create 3 separated PR, there are too many things here. Can you please create:
- 1 PR for fixing the example (Please don't change the indentation of the props)
- 1 PR for fixing the bug in the `Source` component

Also you can't use `import * as MapboxGL from 'mapbox-gl';` this work for the examples because you build mapbox using webpack but it doesn't work as a library exposed to the users. Please only update dependencies if you want in a third PR without changing how we use `mapbox-gl` dependency.

lamuertepeluda:
@alex3165 it makes sense. I should've done many small PRs rather than a big one. Also sorry for the indent, it's vscode fault. I will restore the original one.

Not sure what you mean exactly with the `import * as MapboxGL from 'mapbox-gl';` statement.

I didn't think it was on purpose of adding the whole mapbox dist file like `import * as MapboxGl from 'mapbox-gl/dist/mapbox-gl';` but this clearly imports the minified mapbox which didn't seem correct to me.

In fact I found out this while importing classes like LngLat from mapbox in a project and not being able to compare them with the returned objects from the react-mapbox-gl map (e.g. using `instanceof` operator).

Perhaps there is a different way of [declaring externals](https://webpack.js.org/configuration/externals/) in ts like in webpack, so at least the user of the library can choose?

lamuertepeluda:
I'll try to split this PR

alex3165:
Thank you @lamuertepeluda

lamuertepeluda:
@alex3165 please see PRs #171 , #172 and #173.

Once (if) accepted, I will also make a 4th PR with the additional example (included in this PR) that leverages source-level clustering (i.e. Source and Layer) instead of Cluster + Marker. While it seems way more performant to me, this is also a more accurate porting [of the original Mapbox example](https://www.mapbox.com/mapbox-gl-js/example/cluster/). It requires #171 in order to work properly (without workarounds such as having a setTimeout in onSourceAdded).

Issue #153, opened 2017-03-14
Watch for `layerOptions.filter` change in props change


nrako:
I've noticed that the the `<Layer/>` component doesn't watch `layerOptions` changes, more specifically for `filter` changes. https://github.com/alex3165/react-mapbox-gl/blob/c621f8915cbb7f81b8a7d060f0c5305d548a04fb/src/layer.ts#L180

mapbox api does have a `setFilter` method https://www.mapbox.com/mapbox-gl-js/api/#Map#setFilter

For my use case, it would be very helpful to change the `filter` of a layer, in order to filter what's displayed without having to change the source. Is there any technical reason which would prevent that to be implemented?

alex3165:
Hey @nrako There is no technical reason why it is not implemented, that's something that need to be done. I can jump on it when I have some free time or any PR for this is welcome.

nrako:
Hey @alex3165, thanks for the quick feedback. I might need it soon, I might start a PR, I'll let you know.

nrako:
@alex3165 I started to look into this but I run into an issue with TS, the `@types/mabbox-gl` is wrong:
This line https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/mapbox-gl/index.d.ts#L73
Should be something akin to:
`setFilter(layer: string, filter?: any[]): this;`
The filter argument should be nullable.

I'm not very familiar with TS, any way to circumvent that error type? Or must `@types/mapbox-gl` be fixed first?

alex3165:
Why do you want the filter to be optional ? do you need to call `setFilter` without passing a filter ?

if your problem is that your filter array is of type `filter: string[] | undefined` you should make sure it is not undefined and pass it to `setFilter('layer_string', filter)` you can also force the typescript compiler and tell him it will never be undefined by doing `setFilter('layer_string', filter!)`.

Worse case scenario you can bypass the typings by doing `(map as any).setFilter(layer, filter)`

nrako:
Good question, actually to remove an existing filter I'd simply do `map.setFilter('layer-id', null)`, which works. 
But looking at https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/map.js#L885 make me wonder if there's an more approriate way?

alex3165:
What about defaulting to empty array in order to remove the filter ?

nrako:
Not allowed:
```
Error: layers.layer-10.filter: filter array must have at least 1 element
[...]

PR #169, opened 2017-03-28
Watch for `layerOptions.filter` change in props change

cmrtn:
#153

### DIFF
--- a/src/layer.ts
+++ b/src/layer.ts
@@ -178,7 +178,7 @@ export default class Layer extends React.Component<Props, void> {
   }
 
   public componentWillReceiveProps(props: Props) {
-    const { paint, layout,  before } = this.props;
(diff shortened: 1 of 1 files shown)


src/layer.ts
```diff
@@ -197,6 +197,10 @@ export default class Layer extends React.Component<Props, void> {
       });
     }
 
+    if (!isEqual(props.layerOptions ? props.layerOptions.filter : null, layerOptions ? layerOptions.filter : null)) {
```
alex3165: Can you do something like this instead:
```
    if ((props.layerOptions && layerOptions) && !isEqual(props.layerOptions.filter, layerOptions.filter)) {
      (map as any).setFilter(this.id, props.layerOptions.filter);
    }
```

PR #171, opened 2017-04-03
Fix bug in source data not being loaded by geojson source

lamuertepeluda:
When adding the Source, data don't get loaded.

This fix this behavior by adding a one-time listener to [map:load](https://www.mapbox.com/mapbox-gl-js/api/#map.event:load) and setting source data.

This guarantees data being available after the load event has occurred. 

This PR also adds a `onSourceLoaded` which is similar to `onSourceAdded` and gets called after the map:load event.

I also updated the API.md docs

### DIFF
--- a/src/source.ts
+++ b/src/source.ts
@@ -47,11 +47,11 @@ export default class Source extends React.Component<Props, void> {
     const source = map.getSource(this.props.id) as GeoJSONSource;
 
     const { onSourceLoaded } = this.props;
-    if (onSourceLoaded) {
+    if (source && onSourceLoaded) {
       onSourceLoaded(source);
     }
     // Will fix datasource being empty
-    if (this.props.geoJsonSource && this.props.geoJsonSource.data) {
+    if (source && this.props.geoJsonSource && this.props.geoJsonSource.data) {
       source.setData(this.props.geoJsonSource.data as SourceOptionData);
     }
     map.off('load', this.onData);
--- a/src/source.ts
+++ b/src/source.ts
@@ -15,6 +15,7 @@ export interface Props {
   geoJsonSource?: GeoJSONSourceRaw;
   tileJsonSource?: TilesJson;
   onSourceAdded?: (source: GeoJSONSource | TilesJson) => any;
+  onSourceLoaded?: (source: GeoJSONSource | TilesJson) => any;
 }
 
 export default class Source extends React.Component<Props, void> {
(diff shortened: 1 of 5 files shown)


src/source.ts
```diff
@@ -32,13 +33,30 @@ export default class Source extends React.Component<Props, void> {
 
     if (!map.getSource(this.id) && (geoJsonSource || tileJsonSource)) {
       map.addSource(this.id, (geoJsonSource || tileJsonSource) as any);
+      map.on('load', this.onData);
```
alex3165: don't you just want to listen the load of a single source only ? (the one just above)

lamuertepeluda: I tried, but it didn't seem to update the data.

lamuertepeluda: I mean, there could be some issue with mapbox events (from their side), but the data is never available until map.on('load').

After that, you'll find source._data contains the actual data and is therefore not empty.

There are some related issues on mapbox (some of which are claimed as closed):

* https://github.com/mapbox/mapbox-gl-js/issues/3970
* https://github.com/mapbox/mapbox-gl-js/issues/3958

I tried to wait for `source.load` but it is like it never happens.

lamuertepeluda: I also tried to write a test, but I'm having an hard time with jest 😭 

```tsx
describe('Source', () => {
  let SourceWithContext: any;
  let addLayerMock = jest.fn();
  let addSourceMock = jest.fn();
  let setDataMock = jest.fn();
  let children: any[];
  let childrenWithOneFeature: any[];
  let feature: Object;

  beforeEach(() => {
    addLayerMock = jest.fn();
    addSourceMock = jest.fn();
    setDataMock = jest.fn();
    feature = { coordinates: [-123, 45] };
    children = [{ props: {}}];
    childrenWithOneFeature = [{ props: feature }];

    SourceWithContext = withContext({
      map: React.PropTypes.object
    }, () => ({
      map: {
        addSource: addSourceMock,
        addLayer: addLayerMock,
        on: jest.fn(),
        getSource: jest.fn().mockReturnValue({ setData: setDataMock })
      }
    }))(Source);
  });

  it('Should render source with default options', () => {
    mount(
      <SourceWithContext
        children={children}
      /> as React.ReactElement<any>
    );

    expect(addLayerMock.mock.calls[0]).toEqual([{
      id: 'source-1',
      source: 'source-1',
      type: 'symbol'
    }, undefined]);
  });
...
```

Output

```bash
Source › Should render source with default options

    expect(received).toEqual(expected)

    Expected value to equal:
      [{"id": "source-1", "source": "source-1", "type": "symbol"}, undefined]
    Received:
      undefined

    Difference:

      Comparing two different types of values. Expected array but received undefined.
```

I understand I copied most of the stuff from layer.test.tsx, but I didn't expect it to return always undefined... I don't exactly get how to pass mocked data

lamuertepeluda:
@alex3165 I added a basic test for source. It will not highlight the problem that I've fixed, though, since map is mocked (like for layer test). Anyway it's a beginning...

src/source.ts
```diff
@@ -32,13 +33,30 @@ export default class Source extends React.Component<Props, void> {
...
+    const { map } = this.context;
+
+    const source = map.getSource(this.props.id) as GeoJSONSource;
+
+    const { onSourceLoaded } = this.props;
+    if (onSourceLoaded) {
+      onSourceLoaded(source);
+    }
+    // Will fix datasource being empty
+    if (this.props.geoJsonSource && this.props.geoJsonSource.data) {
```
alex3165: Can you please add a check for source here, as the event could be fired by another component that add a source to the map, which could eventually result to trying to access `setData` of `undefined`

alex3165:
@lamuertepeluda Good job for the test 👍, there is one last thing to add I think and it should be good to go

lamuertepeluda:
@alex3165 you were right, source could be undefined. Now I added the check.

PR #173, opened 2017-04-03
Deps update

lamuertepeluda:
This PR will:  

* Update *all dependencies* (including devDependencies, peerDependencies, example) to their latest versions (I used `npm outdated`).
* add .vscode to .gitignore (local workspace settings, for my mental sanity 😌  )
* Also fix a couple of tslint errors (extra semicolons etc) which were not found by the previous version of tslint.
* Use latest tsc which requires expliciting "libs" in the tsconfig.json
* Finally, fix import of `'mapbox-gl/dist/mapbox-gl'` to just `'mapbox-gl'`. This will works just the same way, because mapbox main module is 'mapbox-gl/dist/mapbox-gl' (see their package.json). **But this will allow people to define alias** in their webpack config files in their projects, e.g. for using mapbox-gl-dev or for compiling it from source. This makes the current library more agnostic on the developer choices on mapbox (e.g. using a debug environment, or compiling it from sources), while remaining compliant to its types thanks to `@types/mapbox-gl`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,31 +53,32 @@
   "homepage": "https://github.com/alex3165/react-mapbox-gl#readme",
   "dependencies": {
     "deep-equal": "^1.0.1",
-    "mapbox-gl": "^0.32.1",
     "reduce-object": "^0.1.3",
-    "supercluster": "^2.2.0"
+    "supercluster": "^2.3.0"
   },
   "peerDependencies": {
-    "react": "^15.0.1",
-    "react-dom": "^15.0.1"
+    "mapbox-gl": "^0.34.0",
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2"
   },
   "devDependencies": {
-    "@types/core-js": "=0.9.36",
-    "@types/enzyme": "^2.7.4",
-    "@types/jest": "^18.1.1",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -53,31 +53,32 @@
...
     "deep-equal": "^1.0.1",
-    "mapbox-gl": "^0.32.1",
     "reduce-object": "^0.1.3",
-    "supercluster": "^2.2.0"
+    "supercluster": "^2.3.0"
   },
   "peerDependencies": {
-    "react": "^15.0.1",
-    "react-dom": "^15.0.1"
+    "mapbox-gl": "^0.34.0",
```
alex3165: Nice I didn't know they changed the `package.json` to target the bundle of mapbox-gl instead of the source files 👍

package.json
```diff
@@ -53,31 +53,32 @@
...
-    "typescript": "^2.1.5"
+    "react-dom": "^15.4.2",
+    "recompose": "^0.22.0",
+    "ts-jest": "^19.0.6",
+    "tslint": "^5.0.0",
+    "tslint-react": "^2.5.0",
+    "typescript": "^2.2.2"
   }
-}
+}
```
alex3165: eof

PR #174, opened 2017-04-03
Update examples / few cleanups

alex3165:


src/map.tsx
```diff
@@ -272,10 +270,6 @@ export default class ReactMapboxGl extends React.Component<Props & Events, State
       });
     }
 
-    if (!isEqual(this.props.style, nextProps.style)) {
-      map.setStyle(nextProps.style);
-    }
```
z0d14c: I'm not 100% sure but I think this might be a breaking change for those of us who had workarounds to remove layers before triggering map style change. I think the correct fix could be to save all the currently added layers, remove them, change the style, then add them back.

Issue #135, opened 2017-02-23
Remove all dependence to 'deep-equal' in favour of reference equality


alex3165:
This might include lot of breaks with the previous version of `react-mapbox-gl` as the map `style` and the Layer `layout` and `paint` properties would be updated checking reference and not deep equality but it would also add lot of performance improvements. @philpl What do you think about it ?

alex3165:
Not sure for the `layout` and `paint` properties actually because we still need to calc the diff to know which key of the `layout` prop to update

lamuertepeluda:
Digging into mapbox code, I found out that they already perform some deep equality checks in order to avoid re-rendering, for instance:

https://github.com/mapbox/mapbox-gl-js/blob/5886421ea9a5ae26797c950ec698be473fe87d17/src/style/style.js#L632

That should be considered in order to avoid unnecessary and expensive deepEquality checks.

You could consider also to use the very efficient [lodash.isEqual](https://www.npmjs.com/package/lodash.isequal), which compares almost anything but functions (that should be enough for Mapbox objects since they are POJOs), or use ImmutableJS which as [its own mechanism](https://facebook.github.io/immutable-js/docs/#/is) for comparison.

PR #179, opened 2017-04-05
Fix component unmounting and map event handling; Add source clustering example

### LINKED ISSUE #178: Map unmounting done wrong?
Hi,

I think [this line of map.tsx](https://github.com/alex3165/react-mapbox-gl/blob/master/src/map.tsx#L198) causes a lot of troubles.

According to [MapBox doc](https://www.mapbox.com/mapbox-gl-js/api/#map#remove) that should not be necessary, implying that all the children should care of releasing only their non-mapbox related resources.

It should just call remove. In the children, we may listen to the [remove event](https://www.mapbox.com/mapbox-gl-js/api/#map.event:remove) and, more important, in the componentWillUnmount we should check if `map.getStyle()` returns undefined before removing sources, layers or whatever.

I'm working on a PR.

lamuertepeluda:
Should fix #178 .

It also adds checks in children component before removing sources or layers.

Source: now it listens to the correct event `sourcedata` like in this [snippet from mapbox tests](https://github.com/mapbox/mapbox-gl-js/blob/0f065bfef21dbe8c3d8e2b4c2b1a5d716eda931e/test/unit/ui/control/logo.test.js#L80)

Layer: improved handling the different cases of map adding/removal, adding some checks on layer or source existence.

Finally, this PR fixes an issue of the style-update example where setState can be called after component unmounting and adds the source-clustering example


src/map.tsx
```diff
@@ -195,13 +195,16 @@ export default class ReactMapboxGl extends React.Component<Props & Events, State
     const { map } = this.state as State;
 
     if (map) {
+      // https://www.mapbox.com/mapbox-gl-js/api/#map#remove
+      // cleans everything including DOM elements, event bindings, web workers, and WebGL resources.
+      map.remove();
       // Remove all events attached to the map
-      map.off();
+      // map.off();
```
alex3165: Remove unnecessary commented lines

src/source.ts
```diff
@@ -27,68 +28,99 @@ export default class Source extends React.Component<Props, void> {
...
+        if (onSourceAdded) {
+          onSourceAdded(map.getSource(id) as GeoJSONSource | TilesJson);
+        }
+      }
     }
   }
 
-  public componentWillReceiveProps(props: Props) {
-    const { id } = this;
+  public shouldComponentUpdate(nextProps: any) {
```
alex3165: I think shouldComponentUpdate is not necessary, I am trying to move away from doing deep equality check on `shouldComponentUpdate` because most of the time we don't need it

src/source.ts
```diff
@@ -27,68 +28,99 @@ export default class Source extends React.Component<Props, void> {
...
-
-      if (onSourceAdded) {
-        onSourceAdded(map.getSource(this.id) as GeoJSONSource | TilesJson);
+    const { id, props } = this;
+    if (e.isSourceLoaded && e.sourceId === this.props.id) {
+      // Refresh data (sometimes unloade without apparent reason)
+      const source = map.getSource(id) as GeoJSONSource | TilesJson;
+      if (source && props.geoJsonSource && props.geoJsonSource.data) {
+        const geoJSONSource = source as GeoJSONSource;
+        geoJSONSource.setData(props.geoJsonSource.data);
```
alex3165: you don't need to do `const geoJSONSource = source as GeoJSONSource;` you can just do `(source as GeoJSONSource).setData`

src/source.ts
```diff
@@ -27,68 +28,99 @@ export default class Source extends React.Component<Props, void> {
...
       const source = map.getSource(id) as GeoJSONSource;
-      source.setData(props.geoJsonSource.data as SourceOptionData);
+      source.setData(prevProps.geoJsonSource.data as SourceOptionData);
+    }
+
+  }
+
+  public componentWillUnmount() {
+    const { map } = this.context;
+    if (map && map.getStyle()) {
```
alex3165: why do you need to check `map.getStyle()` here ?

alex3165:
Thanks @lamuertepeluda for the PR, I had a first look and will check in details later 👍

src/source.ts
lamuertepeluda: Sometimes it returns null. That happens exactly after you call map.remove().

src/source.ts
lamuertepeluda: Allright with that, because these components don't actually render. May be we can clean shouldComponentUpdate lately?

src/source.ts
lamuertepeluda: ok I'm still fighting with tslint errors in vscode 😄

lamuertepeluda:
I fixed a couple of things following your suggestions.

src/source.ts
lamuertepeluda: I tried removing `shouldComponentUpdate` and use `componentWillReceiveProps` instead, but that wouldn't work.

If I do, I keep getting errors like this in the "Style Update" example:
`The layer 'position-marker' does not exist in the map's style and cannot be queried for features.`

Perhaps there are other ways of getting rid of deepEqual as dependency. I'd move this to a different optimizations/enhancement task, such as solving #135

lamuertepeluda:
@alex3165 I aligned the code with the latest change in master.

src/source.ts
lamuertepeluda: Alright, I managed to remove the `shouldComponentUpdate` method and restore the `componentWillReceiveProps`, removing also `deepEqual` from the imports.

It only needed a `map.isLoaded()` guard in order of not throwing the error mentioned above (that probably occurs because when the map is loading it should not update the source).

src/layer.ts
```diff
@@ -154,9 +161,38 @@ export default class Layer extends React.Component<Props, void> {
 
     if (!sourceId) {
       map.addSource(id, source);
+      this.addLayer(map, layer, before);
     }
+    if (sourceId && !map.getSource(sourceId)) {
+      const onSourceLoaded = (e: any) => {
```
alex3165: Can you explain me why you remove the layer and add it back here ?

alex3165:
`src/util/mapbox-type.d.ts` shouldn't be removed

lamuertepeluda:
@alex3165 
I removed `src/util/mapbox-type.d.ts` cause I get compiler errors (duplicate definitions) since I'm using `"@types/mapbox-gl": "^0.30.0"` in the devDependencies

package.json
```diff
@@ -52,30 +52,34 @@
...
-    "react": "^15.4.2",
-    "react-dom": "^15.4.2"
-  },
   "devDependencies": {
     "@types/core-js": "^0.9.41",
+    "@types/deep-equal": "^1.0.0",
     "@types/enzyme": "^2.7.7",
     "@types/geojson": "^1.0.1",
     "@types/jest": "^19.2.2",
+    "@types/mapbox-gl": "^0.30.0",
```
alex3165: remove the types here, I moved it to the local repository so that I can add fixes without having to go through definitelyTyped

src/layer.ts
lamuertepeluda: Method `initialize` gets called in `componentWillMount` (first time, so it cannot be removed) and in `private onStyleDataChange`, that should add and remove the layer in order to update it.

package.json
lamuertepeluda: So do you prefer to have your own mapbox type definitions? Wouldn't it be better to extend the "standard" one?

lamuertepeluda: I compared [your defs](https://raw.githubusercontent.com/alex3165/react-mapbox-gl/master/src/util/mapbox-type.d.ts) and [definitelyTyped](https://raw.githubusercontent.com/DefinitelyTyped/DefinitelyTyped/master/types/mapbox-gl/index.d.ts):  
they are almost identical, except for just a couple of lines... 

I think you could have a much shorter `mapbox-type.d.ts` by extending types where needed or the namespace...

```ts
declare namespace mapboxgl {
  export class MEvented extends Evented {
    on(type: string, listener: Function): this;
    on(type: string, layerId: string, listener: Function): this;
  }

  export class MMap extends Map {
    setFilter(layer: string, filter?: any[]): this;
  }
}
```

lamuertepeluda: Or even better:

```ts
declare namespace ReactMapboxGL {
  export class Evented extends mapboxgl.Evented {
    on(type: string, listener: Function): this;
    on(type: string, layerId: string, listener: Function): this;
  }

  export class Map extends mapboxgl.Map {
    setFilter(layer: string, filter?: any[]): this;
  }
}
```

alex3165: I think it is a good idea, I tried something like :
```
declare module 'mapbox-gl' {
  export class Evented extends mapboxgl.Evented {
    on(type: string, listener: Function): this;
    on(type: string, layerId: string, listener: Function): this;
  }

  export class Map extends mapboxgl.Map {
    setFilter(layer: string, filter?: any[]): this;
  }
}

```

It could work but I get the error : `Duplicate identifier 'Map'` or `Duplicate identifier 'Evented'`

example/package.json
```diff
@@ -10,10 +10,14 @@
...
   "dependencies": {
+    "deep-equal": "^1.0.1",
     "immutable": "^3.8.1",
+    "mapbox-gl": "^0.35.0",
+    "reduce-object": "^0.1.3",
+    "supercluster": "^2.3.0",
     "react": "^15.5.3",
     "react-dom": "^15.5.3",
-    "react-mapbox-gl": "^1.7.0",
+    "react-mapbox-gl": "..",
```
alex3165: can you keep `^1.7.0` (it is used for automatic deployment of the examples) if you want to setup a dev env you have to link react-mapbox-gl, just run `npm run link`

package.json
lamuertepeluda: I actually implemented it in this PR. You have just to user `ReactMapboxGL.<Class>` instead of `MapboxGl.<Class>`, for instance `map: MapboxGl.Map` becomes `map: ReactMapboxGL.Map`.

It works!

Conversely you cannot extend mapbox-gl namespace members in their same ns, since, they are already exported (export default map).

src/map.tsx
```diff
@@ -203,13 +203,9 @@ export default class ReactMapboxGl extends React.Component<Props & Events, State
...
 
     if (map) {
-      // Remove all events attached to the map
-      map.off();
-
-      // NOTE: We need to defer removing the map to after all children have unmounted
-      setTimeout(() => {
-        map.remove();
-      });
+      // https://www.mapbox.com/mapbox-gl-js/api/#map#remove
```
alex3165: you removed `map.off` is it expected ?

src/geojson-layer.ts
```diff
@@ -100,9 +100,10 @@ export default class GeoJSONLayer extends React.Component<Props, void> {
     const { id, layerIds } = this;
     const { map } = this.context;
 
-    map.removeSource(id);
-
-    layerIds.forEach((lId) => map.removeLayer(lId));
+    if (map && map.getStyle()) {
+      map.removeSource(id);
+      layerIds.forEach((lId) => map.removeLayer(lId));
```
alex3165: You can make it shorter :
```
layerIds.forEach(map.removeLayer);
```

src/layer.ts
```diff
@@ -141,7 +141,14 @@ export default class Layer extends React.Component<Props, void> {
     this.hover = hover;
   }
 
-  private initialize = (map: MapboxGL.Map) => {
+  private addLayer = (map: ReactMapboxGL.Map, layer: any, before?: any) => {
+    map.addLayer(layer, before);
+
+    map.on('click', this.onClick);
```
alex3165: I changed this on a recent commit the second argument take the layer id

src/map.tsx
lamuertepeluda: Yep. From [the link](https://www.mapbox.com/mapbox-gl-js/api/#map#remove) I put in the comments:

> Clean up and release all internal resources associated with this map.
> This includes DOM elements, event bindings, web workers, and WebGL resources.

example/package.json
lamuertepeluda: Ok I didn't know that

alex3165:
@lamuertepeluda sorry for all the conflicts I feel like this PR is again a bit to big, we will never manage to merge it if we don't break it into smaller PR, do you mind creating different PR's against the new master ?

Basically:

- One PR for the examples
- One PR for the modification on the Source component
- One PR for the modification on the Layer component

I think you should be careful to have only the required changes for 1 feature / fix for each PR otherwise if there is a breaking changes it is harder to track what introduced it.

lamuertepeluda:
I think you're right. I will open several issues and split this PR into separate branches against current master from which I will open the PR, referencing the related issues.

alex3165:
Thanks 👍

alex3165:
@lamuertepeluda Any update on splitting this PR ? 👍

lamuertepeluda:
@alex3165  sorry I didn't have time to do it yet.

PR #196, opened 2017-04-18
Exposed renderWorldCopies Map option to properties.

lastsys:
Added map property `renderWorldCopies` according to the Mapbox GL API (https://www.mapbox.com/mapbox-gl-js/api/).

My editor also killed some trailing whitespace automatically in the files I touched. Hope this is ok.


alex3165:
LGTM 👍

PR #203, opened 2017-04-26
Zoom control buttons should not cause form submission

justin-hackin:
add  type="button" to prevent default behaviour (form submission inside form)


alex3165:
Thanks for the PR 👍

alex3165:
It is published under `1.9.2` tag :)

PR #207, opened 2017-05-01
Examples with custom tiles

stereobooster:
Examples with custom tiles. Probably need to clean up it a bit. What do you think?


stereobooster:
tried react-snapshot, but it has this issue https://github.com/geelen/react-snapshot/issues/16

stereobooster:
Added Marker, Popup example

alex3165:
@stereobooster Thanks for the PR, Great job but I am not sure about the value of having examples with custom tiles. In the end what changed for the library is just the style passed down to the map, it doesn't show more features that `react-mapbox-gl` provide.

What I can propose you is to keep this into a different repository and I will add a link to it in the readme as "More examples using react-mapbox-gl", what do you think about it ?

stereobooster:
My intention was to show that library actually not coupled with Mapbox tile server on one side, on other side this is not well documented feature. 

I like how examples organised in [styleguidist](http://react-styleguidist.js.org/) or in [https://uber.github.io/react-map-gl/#/](react-map-gl): Code example + Demo.

From my POV the more examples the better. WDYT?

alex3165:
I agree with you for the organisation of the examples (code + example), let me have a look at the actual result of the examples on your PR (will check this week).

stereobooster:
Just a note this is not deployable without some kind of static prerenderer, like react-snapshot, because I use react-router.

alex3165:
Sorry I will close this PR as the examples changed completely with the version 2.0.0

Issue #204, opened 2017-04-27
NavigationControl


alpha-beta-soup:
Is it possible to include a [NavigationControl](https://www.mapbox.com/mapbox-gl-js/example/navigation/) component? Currently when the map is rotated or tilted, it is difficult to return back to north-up and straight-down. With the standard Mapbox GL JS controls, you can just click on the compass icon (navigation control) to reset the view at the current zoom. It's possible to workaround with `map.addControl(new MapboxGl.NavigationControl())`, except that it conflicts with the `ZoomControl` component. I note that there's a [`NavigationControl`](https://github.com/alex3165/react-mapbox-gl/blob/463d9c2d059d28d0eb785b79d615cb144a80733d/src/util/mapbox-type.d.ts#L353) which I assumed I'd use in exactly the same way as `ZoomControl` or `ScaleControl` (just with no props) but it doesn't render, and it's not used in any example. So is it possible to include a `NavigationControl`?

alex3165:
There is currently no Navigation control component, I would be happy to see such a component. Although I would use the same approach as `ZoomControl` and `ScaleControl` component and create real React component that reset the pitch and rotation of the map to their initial state manually.

PR #227, opened 2017-05-21
Rotation control component

fgkramer:
I added the Rotation Control Widget which was requested on #204, but before it's merged, I want to check if making a `NavigationControl` wrapper would be the best solution when the user wants to display the `ZoomControl` component and the `RotationControl` at the same time.

Having a `NavigationControl` component would make the `hover` property in the state easier to manage between the navigation components and also make the styling easier, as right now the borders of the component this PR collide with the `ZoomControl` ones.

Shortcomings of this PR:
- No tests yet (will wait until the issue with the wrapper is defined).
- Displaying both `ZoomControl` and `RotationControl` doesn't look great aesthetically.
- Styling needs some polishing.
- I used the `transform` property of the map object though it's not exposed on the mapbox-gl typings.

Doubts regarding guidelines:
- I used the compass icon that mapbox-gl itself uses on it's css, I don't know if this is desired.

If any maintainer wants to discuss approaches I'd be happy to help.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,6 +10,7 @@ import ScaleControl from './scale-control';
 import Marker from './marker';
 import Source from './source';
 import Cluster from './cluster';
+import RotationControl from './rotation-control'
 
 injectCSS(window);
 
@@ -23,7 +24,8 @@ export {
   ScaleControl,
   Marker,
   Source,
-  Cluster
+  Cluster,
+  RotationControl
 };
 
 export default Map;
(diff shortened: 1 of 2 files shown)


alex3165:
HI @fgkramer, Thanks for your submission. I have reviewed some bit of your code and pushed a new PR, could you check it and tell me if you are fine with it https://github.com/alex3165/react-mapbox-gl/pull/235 ?

PR #232, opened 2017-05-25
Include new and removed keys in diff results

cmrtn:
The current diff only returns keys that are in both objects, so anything added or removed from paint/layout doesn't get updated on the map.


alex3165:
Thank you very much @cmrtn, could you add some unit test for this new diff function?

alex3165:
Awesome thank you! 👍

PR #235, opened 2017-05-26
Rotation control component

alex3165:
Should resolve: https://github.com/alex3165/react-mapbox-gl/issues/204


alpha-beta-soup:
Is it possible to make the `onClickCompass` reset the map pitch, too?

PR #250, opened 2017-06-04
Add none pointerEvents when scrollZoom is true

alex3165:


stereobooster:
What about standalone markers?

alex3165:
They use `src/projected-layer.tsx` under the hood

PR #273, opened 2017-06-18
removing empty accessToken=

jpdevries:
doesn't seem needed now that it is set on the Factory. Also invalid JSX.


alex3165:
Thanks 👍

Issue #246, opened 2017-06-03
How to get reference to map object?


stereobooster:
This doesn't seem to work

```js
ref={ map => { this.map = map } }
```

I want to do something like [this](https://www.mapbox.com/mapbox-gl-js/example/scroll-fly-to/), but I need reference for this.

aleksplus:
did you try access map object from state `this.map.state.map`?

refs should work after component is [mounted](https://facebook.github.io/react/docs/refs-and-the-dom.html#adding-a-ref-to-a-dom-element)

stereobooster:
I tried this, but unfortunately it is null 

```js
  componentDidMount() {
    console.log(this.map.state.map)
  }
```

strange thing if I log `this.map` I see `state.map` WAT? Still looking into this

stereobooster:
```js
class ReactMapboxGlCustom extends ReactMapboxGl {
  componentDidMount() {
    ReactMapboxGl.prototype.componentDidMount.call(this)
    this.props.onMount && this.props.onMount(this.state.map) // this.state.map is null
  }
}
```

stereobooster:
For future googlers there is `onStyleLoad` which is called when `map` is available

```jsx
<ReactMapboxGl onStyleLoad={ (map) => { ... } } />
```

PR #274, opened 2017-06-18
onLoad callback

stereobooster:
This PR addresses problem of getting reference to map object. See #246

Not sure this is the right way to do it. Maybe there is better option, but haven't found reliable way to get reference to the map.


stereobooster:
Wait a second. `onStyleLoad` is what I was looking for

Issue #167, opened 2017-03-27
Cluster component not updating if child Markers change


drekli:
Thought I'd try out the new Marker and Cluster components (awesome idea!), and noticed that whilst the child Markers show/hide on a query change, the Clusters always remain the same from first load.

Same pointCount and coords, even if the children underneath change or no longer exist. Also means no new Clusters appear for Markers that didn't exist on first load and start with a parent Cluster (or were outside of the getBounds on first load). Tried out the demo and but the data doesn't change so couldn't reproduce the issue there. Looked at the component as well, and only saw mapChange updating the pointCount, but not for a new query.

Seeing as the child Markers update happily along with other map components, I couldn't think what the issue with Clusters not updating.

The chance of html Markers makes me not want to switch back to Layers, so any help is appreciated!

alex3165:
@drekli Can you try out with react-mapbox-gl v2?

drekli:
Sure thing, I'll look forward to testing that! Thanks for the heads up on v2 and all the hard work!

marek-sed:
this is still happening even on version 2

alex3165:
@marek-sed, @Wykks will look into that

marek-sed:
@Wykks I looked at the code I think I found the issue and was able to fix it properly.
Hope it helps
#286

alex3165:
Released with version `2.0.6`

PR #286, opened 2017-06-26
Issue #167 - reload features on when children changed

marek-sed:
Reloading features on componentWillReceiveProps


Wykks:
Thanks ! That's almost perfect.
But, there is a small issue, when `mapChange` is called (inside `componentWillReceiveProps`) it won't always call `this.setState({ clusterPoints: newPoints });` (when the visible features are the same). But since the features are recreated (so the weakmap), in this case `this.featureClusterMap.get(feature)` inside render will not work, thus skipping all markers.
(I'll will fix that directly).

PR #290, opened 2017-06-28
Fix #136 fitBounds burst issue

### LINKED ISSUE #136: fitBounds burst issue
Hi,

Did you happen to see that a burst of fitBounds changes would be discarded.

I have the following log from `render()`:

```
bounds: [[-14.956032191681425,27.33871139614223],[-13.151319808318574,29.378776603857773]]

bounds: [[-14.956032191681425,27.33871139614223],[-13.151319808318574,29.378776603857773]]

bounds: [[-14.956032191681425,27.33871139614223],[-13.151319808318574,29.378776603857773]]

bounds: [[-2.8241818150222513,47.1865573870424],[2.5686652150222513,55.8281444129576]]

bounds: [[-2.8241818150222513,47.1865573870424],[2.5686652150222513,55.8281444129576]]
```

What happens is that location remains at `[[-14.956032191681425,27.33871139614223],[-13.151319808318574,29.378776603857773]]` and does not change to `[[-2.8241818150222513,47.1865573870424],[2.5686652150222513,55.8281444129576]]`

Changing bounds manually from console worked though, i.e:

```js
map.state.map.fitBounds([[-2.8241818150222513,47.1865573870424],[2.5686652150222513,55.8281444129576]], { offset: [0, -100]})
```

My code is fairly simple:

```
const mapBounds = this.toLngLatBounds(bounds);
const mapBoundsOptions = { offset: [0, -100] };

<ReactMapboxGl 
  style="mapbox://styles/..."
  accessToken={ accessToken }
  containerStyle={{ height: '100%' }} 
  interactive={ false }
  fitBounds={ mapBounds }
  fitBoundsOptions={ mapBoundsOptions }>
```

I'll try to come up with isolated test case.

gschjetne:
This should solve #136.

A reference to the map object needs to be available at all times, or it won't receive the updated `fitBounds` (or any other property for that matter).

### DIFF
--- a/src/map.tsx
+++ b/src/map.tsx
@@ -273,16 +273,13 @@ const ReactMapboxFactory = ({
       };
 
       const map = new MapboxGl.Map(opts);
+      this.setState({ map });
 
       if (fitBounds) {
         map.fitBounds(fitBounds, fitBoundsOptions);
       }
 
       map.on('load', (evt: React.SyntheticEvent<any>) => {
-        if (this._isMounted) {
-          this.setState({ map });
-        }
-
         if (onStyleLoad) {
           onStyleLoad(map, evt);
         }
--- a/src/map.tsx
+++ b/src/map.tsx
@@ -125,6 +125,7 @@ export interface Props {
 
 export interface State {
   map?: MapboxGl.Map;
+  ready: boolean
 }
 
 // Static Properties of the map
@@ -209,7 +210,8 @@ const ReactMapboxFactory = ({
     };
 
     public state = {
-      map: undefined
+      map: undefined,
+      ready: false
     };
 
     // tslint:disable-next-line:variable-name
@@ -273,13 +275,17 @@ const ReactMapboxFactory = ({
       };
 
       const map = new MapboxGl.Map(opts);
-      this.setState({ map });
+      this.setState({ ...this.state, map });
 
       if (fitBounds) {
         map.fitBounds(fitBounds, fitBoundsOptions);
       }
 
       map.on('load', (evt: React.SyntheticEvent<any>) => {
+        if (this._isMounted) {
+          this.setState({ ...this.state, ready: true })
+        }
+
         if (onStyleLoad) {
           onStyleLoad(map, evt);
         }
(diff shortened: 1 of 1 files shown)


gschjetne:
@alex3165 Do you have some time to take a look at this?

alex3165:
Hey @gschjetne, thanks for reporting but we need to wait until the style of the map is loaded before rendering any children of the Map component (Layer, Source...). Otherwise nothing is displayed until you trigger a re-rendering the Map. Maybe we can set the reference of the map as you did and use another property in the component state to render the children elements when the style is loaded. In any case you can run the examples to make sure you changes doesn't break anything.

gschjetne:
Would this do the trick??

alex3165:
That works well, could we add a test for the original issue? to make sure we don't introduce regression later.

gschjetne:
Good idea. I'll give it a crack tomorrow at the office.

src/map.tsx
```diff
@@ -273,14 +275,15 @@ const ReactMapboxFactory = ({
       };
 
       const map = new MapboxGl.Map(opts);
+      this.setState({ ...this.state, map });
```
alex3165: You don't need to spread `this.state`, `setState` is doing a merge and not a replace

src/map.tsx
```diff
@@ -273,14 +275,15 @@ const ReactMapboxFactory = ({
...
 
       if (fitBounds) {
         map.fitBounds(fitBounds, fitBoundsOptions);
       }
 
       map.on('load', (evt: React.SyntheticEvent<any>) => {
         if (this._isMounted) {
-          this.setState({ map });
+          this.setState({ ...this.state, ready: true })
```
alex3165: same as above

src/map.tsx
alex3165: You don't need to spread `this.state`, `setState` is doing a merge and not a replace

src/map.tsx
alex3165: same as above

alex3165:
I will get this merged, we can write a unit test later ! 👍  thank you for the PR

alex3165:
Released with version `2.1.0` 👍

gschjetne:
Excellent, thanks!

Issue #208, opened 2017-05-02
Cluster click implementation


porchdog:
Hey all.  I'm trying to implement a way to show a popup list of clustered items when a cluster has been clicked.  Problem is, the cluster factory does not get passed the items that make up the cluster.  So when it is clicked, I don't know which items are in that cluster.

Any chance that the API could be modified such that the factory gets passed the array of markers that make up the cluster?  That would continue to provide the number of points (array length) as well as being able to know which markers make up the cluster.

Thanks for the component!

alex3165:
Hey thanks for the issue it would indeed be possible. I can have a look when I have some time or I would be happy to have some help on this.

Wykks:
Hi, I implemented this feature, with minimal changes to the API (no breaking changes).
https://github.com/alex3165/react-mapbox-gl/pull/306

alex3165:
Release with version `2.2.0`, see: https://github.com/alex3165/react-mapbox-gl/releases/tag/v2.2.0

PR #306, opened 2017-07-23
WIP: feat(cluster): add getLeaves argument to factory

Wykks:
Feature asked in #208 
It's working, but something must be changed in the cluster API to transmit original feature properly.
Until today, converting `Marker` props to feature was good enough, since we didn't need to access properties of the original feature.
Right now the only thing `getLeaves` can return is `Marker` component, which doesn't contain original feature data.
So in order to show what's inside the `Marker` we have to do this (two children here to skip the div element): `{leaf.props.children.props.children}`).

We could store the feature inside `Marker` and get it like that: `{leaf.props['data-feature']}`.
But it doesn't feel right...

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -406,7 +406,9 @@ clusterMarker = (coordinates) => (
 ```
 
 ### Properties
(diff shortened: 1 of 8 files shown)


Wykks:
PR updated, with test and docs. Improved the example with `feature` passed as `data-feature`. But not sure if it's a good practice... What do you guys think ?

Issue #326, opened 2017-08-23
[Popup] bug with className. Custom className override defaultClassName


lkazberova:
Hi!

Sorry for my english. 
Use case : set className to Popup, for example "my-popup"
Expected result: defaultClassName concated with custom className, for example "my-popup mapboxgl-popup mapboxgl-popup-anchor-bottom"
Actual result:  "my-popup" without defaultClassName

Why: https://github.com/alex3165/react-mapbox-gl/blob/master/src/popup.tsx#L34
computed className overrided with props.className
`<ProjectedLayer className={className} {...props as any}>`
Should be
`<ProjectedLayer {...props as any} className={className} >`

alex3165:
Hey @lkazberova it was a mistake it should be fixed with version `2.2.1` of the library. Feel free to close this issue if it is good now.

lkazberova:
Ok will do with test for popup

PR #327, opened 2017-08-23
fix issue #326 [Popup] bug with className

lkazberova:
[Popup] bug with className. Custom className override defaultClassName


alex3165:
@lkazberova Very good job 👍  thank you for the PR

Issue #325, opened 2017-08-22
onClick event for GeoJSONLayer?


jussirantala:
In mapbox documentation it seems there's a click event for map layers. I think it should be in the props of GeoJSONLayer component so it could be used.

jussirantala:
Nvm. I can set id for the GeoJSONLayer and use this:
`  map.on('click', 'LayerID', function (e) {
      // do stuff
    });
`

jussirantala:
Oh, the id is set like this:
`private id: string = this.props.id || `geojson-${generateID()}`;`

How I can get the ID which it sets for the layer?

alex3165:
@jussirantala Thank you for the issue and sorry for the late reply. I am happy to have an `onClick` prop on the GeoJson component if anyone want to work on this 👍 . Otherwise I will try to find some times for this asap.

alex3165:
Released with version `2.3.0`

rofrol:
Label `In progress` can be removed

Issue #312, opened 2017-07-31
GeoJSONLayer doesn't update on layerOptions.filter change


tofuness:
Related to #169. `map.setFilter` was only added to `Layer`.

4xDMG:
Is it okay if I work on this issue?

tofuness:
@4xDMG That would be awesome. I am currently using Layer + Source instead.

Wykks:
@4xDMG  Sure any PR for this is welcome :)

4xDMG:
Awesome. I'll give it ago tomorrow.

alex3165:
@4xDMG any update on this by any chance?

SBRK:
Since I need that, I'm also woring on it

alex3165:
Released with version `2.3.0`

PR #332, opened 2017-08-30
Added mouse handlers to GeoJSONLayer

SBRK:
I needed a way to know when mouse was hovering/clicking on GeoJSONLayer features

### DIFF
--- a/src/geojson-layer.ts
+++ b/src/geojson-layer.ts
@@ -167,21 +167,18 @@ export default class GeoJSONLayer extends React.Component<Props, {}> {
 
     map.removeSource(this.id);
     map.off('styledata', this.onStyleDataChange);
-    this.layerIds.forEach(lId => map.removeLayer(lId));
 
-    for (const type in typeToLayerLUT) {
-      if (typeToLayerLUT.hasOwnProperty(type)) {
-        for (const event in eventToHandler) {
-          if (eventToHandler.hasOwnProperty(event)) {
-            const prop = typeToLayerLUT[type] + eventToHandler[event];
+    Object.keys(typeToLayerLUT).forEach(type => {
+      Object.keys(eventToHandler).forEach(event => {
+        const prop = typeToLayerLUT[type] + eventToHandler[event];
 
-            if (this.props[prop]) {
-              map.off(event, this.buildLayerId(type), this.props[prop]);
-            }
-          }
+        if (this.props[prop]) {
+          map.off(event, this.buildLayerId(type), this.props[prop]);
         }
-      }
-    }
+      });
+    });
+
+    this.layerIds.forEach(lId => map.removeLayer(lId));
   }
 
   public componentWillReceiveProps(props: Props) {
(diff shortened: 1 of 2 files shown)


alex3165:
Cool, this PR should fix https://github.com/alex3165/react-mapbox-gl/issues/325, I will take a look at it this evening 👍

src/geojson-layer.ts
```diff
@@ -98,8 +107,26 @@ export default class GeoJSONLayer extends React.Component<Props, {}> {
...
       before
     );
+
+    this.mapLayerMouseHandlers(type)
   };
 
+  private mapLayerMouseHandlers = (type: string) => {
+    const { map } = this.context;
+
+    const layerId = `${this.id}-${type}`;
```
alex3165: Could you create a function `buildLayerId` and use it in line 88 too?

SBRK: done

alex3165:
Hey @SBRK , I had a look at the code and it look good. Could listen off the event on `componentUnmount` ? Probably something like:
```
this.layerIds.forEach(id => {
  Object.keys(eventToHandler).forEach(event => {
    const handler = this.props[`${typeToLayerLUT[type]}${eventToHandler[event]}`] || null;
    if (handler) {
      map.off(event, id);
    }
  });
  map.removeLayer(lId)
});
```

SBRK:
Actually, managing the `off` might be a bit trickier, as it requires to pass the handler. I would also need to manage changing listeners in the `componentWillReceiveProps`, I didn't put time into it yet, as it's not a case I needed to handle for the moment. Will do it during the day if I have time (as well as documentation)

SBRK:
@alex3165 I added the `componentWillReceiveProps` / `componentWillUnmount` code as well as the doc

src/geojson-layer.ts
```diff
@@ -137,6 +168,20 @@ export default class GeoJSONLayer extends React.Component<Props, {}> {
     map.removeSource(this.id);
     map.off('styledata', this.onStyleDataChange);
     this.layerIds.forEach(lId => map.removeLayer(lId));
+
+    for (const type in typeToLayerLUT) {
```
alex3165: Could you use `Object.keys(typeToLayerLUT).forEach ` to keep the coding style consistent with the rest of the codebase? Same for eventToHandler and same for lines 223 to 227

src/geojson-layer.ts
```diff
@@ -169,11 +214,31 @@ export default class GeoJSONLayer extends React.Component<Props, {}> {
...
 
           Object.keys(layoutDiff).forEach(key => {
-            map.setLayoutProperty(`${this.id}-${type}`, key, layoutDiff[key]);
+            map.setLayoutProperty(this.buildLayerId(type), key, layoutDiff[key]);
           });
         }
       }
     }
+
+    for (const type in typeToLayerLUT) {
```
alex3165: Line 209 we already iterate over the `typeToLayerLUT`, if you could insert your block in there that would be better. Also it is a technical debt I would rather use `Object.keys().forEach` too

src/geojson-layer.ts
```diff
@@ -169,11 +214,31 @@ export default class GeoJSONLayer extends React.Component<Props, {}> {
...
+      if (typeToLayerLUT.hasOwnProperty(type)) {
+        for (const event in eventToHandler) {
+          if (eventToHandler.hasOwnProperty(event)) {
+            const prop = typeToLayerLUT[type] + eventToHandler[event];
+
+            if (!isEqual(props[prop], this.props[prop])) {
```
alex3165: To test equality on a function I think you can go for a reference check straight away so `props[prop] === this.props[prop]` no need for value equality and `===` is much faster

SBRK:
@alex3165 all done. I regrouped the change of layout/paint/handlers in a same `Object.keys(typeToLayerLUT).forEach`

alex3165:
Alright, it is looking good, there one more thing missing, we need to add all the events to the typescript interface, I think it would be better to refactor a small bit too:

Could you extract all the props per shape type from the global Props interface (example for line):
```
export interface LineProps {
  lineLayout?: MapboxGL.LineLayout;
  linePaint?: MapboxGL.LinePaint;
  lineOnMouseDown?: MouseEvent;
  lineOnMouseUp?: MouseEvent;
  lineOnMouseMove?: MouseEvent;
  lineOnMouseEnter?: MouseEvent;
  lineOnMouseLeave?: MouseEvent;
  lineOnClick?: MouseEvent;
}
```

Where MapEvent is:
```
export type MouseEvent = (evt: Event) => any
```

In the end you can compose everything back to the global `Props` by doing:
```
export interface Props extends LineProps, ... {
...
}
```

If you don't feel confortable with those changes I can add them once we merge this PR it is fine

SBRK:
I'm not familiar with TypeScript so I think it's better if you do it :)

alex3165:
Alright, it look good to me I will add the missing stuff (TS, Tests, Example) after the merge, everything good for you @SBRK ?

alex3165:
For history, this PR should also solve https://github.com/alex3165/react-mapbox-gl/issues/312

Issue #333, opened 2017-08-31
Feature request: zoom to bounds of cluster on click


rhewitt22:
When the user clicks on a cluster the map zooms to the bounds of the cluster.

Similar to how [leaflet.markercluster](https://github.com/Leaflet/Leaflet.markercluster) works.

[Example](https://leaflet.github.io/Leaflet.markercluster/example/marker-clustering-realworld.388.html).

alex3165:
Hey @rhewitt22 , I think that could be possible, how would you see it in term of API? 

What about something like:
```
<Cluster ClusterMarkerFactory={...} zoomOnClick={true}>
...
</Cluster>
```

cc @Wykks

rhewitt22:
@alex3165 it doesn't get any simpler of an API than that; I like it.
😃

alex3165:
Merged and released with version `2.5.0`

PR #336, opened 2017-09-03
Feature/cluster zoom

Wykks:
#333 

I'll probably add the hover effect like in leaflet.markercluster next time.

### DIFF
--- a/src/cluster.tsx
+++ b/src/cluster.tsx
@@ -1,13 +1,13 @@
 import * as React from 'react';
 const PropTypes = require('prop-types'); // tslint:disable-line
 
-import { Map } from 'mapbox-gl';
+import { Map, LngLatBounds } from 'mapbox-gl';
 import { Props as MarkerProps } from './marker';
 const supercluster = require('supercluster'); // tslint:disable-line
 import * as GeoJSON from 'geojson';
 import { Feature } from './util/types';
 import * as bbox from '@turf/bbox';
-import { polygon } from '@turf/helpers';
+import { polygon, featureCollection } from '@turf/helpers';
 
 export interface Props {
   ClusterMarkerFactory(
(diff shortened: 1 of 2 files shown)


src/cluster.tsx
```diff
@@ -150,12 +154,46 @@ export default class Cluster extends React.Component<Props, State> {
...
+    // https://github.com/mapbox/mapbox-gl-js/issues/5249
+    this.context.map.fitBounds(LngLatBounds.convert(childrenBbox as any), {
+      padding: this.props.zoomOnClickPadding!
+    });
+  };
+
+  private findMarkerElement(
+    target: HTMLElement,
+    element: HTMLElement
+  ): HTMLElement {
```
alex3165: Return type should be inferred no?

src/cluster.tsx
```diff
@@ -150,12 +154,46 @@ export default class Cluster extends React.Component<Props, State> {
       .map((leave: Feature) => this.featureClusterMap.get(leave));
   };
 
+  public zoomToClusterBounds = (event: React.MouseEvent<HTMLElement>) => {
```
alex3165: Destruct event? `{ currentTarget, target }`

src/cluster.tsx
Wykks: 'findMarkerElement' implicitly has return type 'any' because it does not have a return type annotation and is referenced directly or indirectly in one of its return expressions.
:)

src/cluster.tsx
Wykks: Yeah, but it's not very convenient to type `{ currentTarget, target }` instead of `event`

alex3165: It is not a big deal, happy with both to be honest

Issue #344, opened 2017-09-09
Support for adding custom image at runtime using map.addImage


alex3165:
Mapbox-gl allow to add custom images at runtime using `loadImage` and `addImage`, I don't think `loadImage` is necessary, instead user should pass an `HTMLImageElement` to `addImage` and query for the image by their own using `new window.Image()`.

In term of API I am not sure what is best between using prop of Layer component:
```
<Layer
  addImage={image}
>
</Layer>
```

Or having a component for that:
```
<ImageSource image={image}/>
```

See: https://www.mapbox.com/mapbox-gl-js/api/#map#addimage

alex3165:
For reference, the prop `images` has been added to react-mapbox-gl with the release `2.5.0`, it can be use this way:
```
const image = new Image(30, 30);
image.src = 'image_source_url_or_base64';

const images = ['myImage', image];
...
<Layer
  layout={{ icon-image: 'myImage' }}
  images={images}
/>
```

redixhumayun:
Hey, when I try this, I get the following error:
`react-dom.development.js:12879 Uncaught DOMException: Failed to execute 'drawImage' on 'CanvasRenderingContext2D': The HTMLImageElement provided is in the 'broken' state.
`
Any idea why?

Valid:
This works for me (shows the image) but gives me a few errors when removing the <Layer> the new image is associated with:

```
const image = new Image(40, 60);
image.src = '/img/new-icon.svg';

const images = ['newIcon', image];

{show && (
<Layer
  type="symbol"
  layout={{ 'icon-image': 'newIcon' }}
  images={images}
>
...
)}
```

Errors when `!show`:
```
mapbox-gl.js:484 Uncaught TypeError: Cannot read property 'style' of undefined
    at e.removeImage (mapbox-gl.js:484)
    at Array.forEach (<anonymous>)
    at Layer../N:/Dropbox/Work/wildcrowd-web-app/node_modules/react-mapbox-gl/lib/layer.js.Layer.componentWillUnmount (layer.js:129)
    at ReactCompositeComponent.js:406
    at measureLifeCyclePerf (ReactCompositeComponent.js:73)
    at ReactCompositeComponentWrapper.unmountComponent (ReactCompositeComponent.js:405)
    at Object.unmountComponent (ReactReconciler.js:76)
    at ReactCompositeComponentWrapper.unmountComponent (ReactCompositeComponent.js:415)
[...]

PR #345, opened 2017-09-11
Support for adding custom images with Layer component

alex3165:
Should solve https://github.com/alex3165/react-mapbox-gl/issues/344


Daryn17:
@alex3165  this pull not solve issues #344

Issue #355, opened 2017-09-20
fitBounds not being updated when changing fitBoundsOptions


rofrol:
```jsx
import React, { Component } from 'react';
import ReactMap, { Marker } from 'react-mapbox-gl';

const accessToken = "pk.eyJ1IjoiYWxleDMxNjUiLCJhIjoiY2o0MHp2cGtiMGFrajMycG5nbzBuY2pjaiJ9.QDApU0XH2v35viSwQuln5w";
const style = "mapbox://styles/mapbox/streets-v9";

const BOUNDS = [
  [88.537838, 26.564062],
  [92.325374, 28.488102]
];

const CENTER = [90.2361995, 27.4128857];

const Map = ReactMap({
  accessToken
});

const mapStyle = {
  height: '100vh',
  width: '100vw'
};

class App extends Component {
  constructor(props) {
    super(props);
    this.state = { fitBoundsOptions: { offset: [640 / 2, 0] }};
    this.setOffset = this.setOffset.bind(this);
  }
  setOffset(e) {
    let offset = [ parseInt(e.target.value, 10) / 2, 0];
    this.setState({ fitBoundsOptions:  { offset } });
  }
  render() {
    return (
      <div>
       Sidebar
       <input type="number" step="2" onChange={this.setOffset} />,
        offset[0]: {this.state.fitBoundsOptions.offset[0]}
       <Map
         style={style}
         containerStyle={mapStyle}
         fitBounds={BOUNDS}
         fitBoundsOptions={this.state.fitBoundsOptions}
       >
         <Marker
           coordinates={CENTER}
         >
           ---------------Center---------------
         </Marker>
       </Map>

      </div>
    );
  }
}

export default App;
```

https://www.webpackbin.com/bins/-KuV-ovZ4mCpiYXzbsnA

rofrol:
I'm using the workaround to recreate the map:

```jsx
  componentWillReceiveProps (nextProps) {
    if (nextProps.sidebarWidth !== this.props.sidebarWidth) {
      // hack to rerender map https://github.com/alex3165/react-mapbox-gl/issues/355
      MapBox = ReactMapboxGl({ accessToken: mapBoxConfig.accessToken });
    }
  }
```

alex3165:
[...]

PR #357, opened 2017-09-20
Update fitboundsOptions even if fitBounds doesn't change

alex3165:
Should fix https://github.com/alex3165/react-mapbox-gl/issues/355


rofrol:
Shouldn't `isEqual` be used?

alex3165:
@rofrol Sorry if I come back to you a bit late, I think we will merge this PR into the v3.0.0 branch as it is a breaking change. Therefore we decided we would go for reference equality to check for updates but actually I think that for this use case it make actually sense to check for a value equality so using `isEqual` would make sense indeed

PR #383, opened 2017-09-26
fix(cluster): actually use the arguemnts provided to getLeaves

hobofan:
Pass through the arguments to supercluster so the pagination actually works [as documented](https://github.com/alex3165/react-mapbox-gl/blob/ac71beba10b644fcfb8b6345ab13008e956dd554/docs/API.md#properties-9)

### DIFF
--- a/src/cluster.tsx
+++ b/src/cluster.tsx
@@ -147,10 +147,10 @@ export default class Cluster extends React.Component<Props, State> {
       return feature;
     });
 
-  private getLeaves = (feature: Feature) => {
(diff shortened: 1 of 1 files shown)


alex3165:
Seems good to me, cf @Wykks

Wykks:
Whoa, how could I miss that... Thanks !

alex3165:
Released with patch version `2.5.1`

hobofan:
Thanks for the quick release! :)

Issue #320, opened 2017-08-12
Loss of existing layers when map style is changed (how to)?


systemlevel:
Not sure if anyone else is experiencing the following issue when map style are changed:

In my case, I'm storing the current mapbox map style in a react state variable. I then setState() to update the map style and it works just fine.

However, the layers I had displayed on the map before changing the style state get removed on the new style.  I have the layer in question loaded into mapbox on both map styles.

Is there a way to preserve the layers that were on the style previous to switching it to another style?  If not, what would be the best way to reapply those layers on the new style?  

The attached image shows the layer that is visible on the outdoors map (black area) while being missing from the satellite style.

This may not be an "issue" with react-mapbox-gl.  However, I'm having a hard time figuring out a solution to making it work.  Whenever I try to reapply the layer to the satellite style I get an error about the styles not being loaded yet.  I have tried to use the callback onStyleLoad in an attempt to reapply the layers at that point but that only fires one time when the outdoors style is initially loaded.

Thank you for this great library.  Really have enjoyed using this.

<img width="515" alt="screen shot 2017-08-11 at 10 02 32 pm" src="https://user-images.githubusercontent.com/894983/29237720-8c511b18-7ee1-11e7-9931-47a8b11b7d6c.png">
<img width="370" alt="screen shot 2017-08-11 at 10 02 47 pm" src="https://user-images.githubusercontent.com/894983/29237721-8c514a8e-7ee1-11e7-9685-0d7cd5ec75c0.png">

alex3165:
Hey @systemlevel , there is possibly some issues with layers being removed when changing the style of the map, can you tell me the component used that doesn't show up when you change the style?

If you could reproduce the issue using [react-mapbox-gl-debug](https://github.com/alex3165/react-mapbox-gl-debug) that would be very helpful for me to debug.

systemlevel:
Hi @alex3165, the layer that gets removed when I change map styles is a tileset I have hosted on mapbox.  

To display this tileset layer on the "outdoors" map style you see above, I create a source and a layer. Again, this is refrencing a map hosted on mapbox referenced by the url below.  Here is what that looks like:

`<Source id="utah_unitsgeojson" tileJsonSource={{ type: 'vector', url: "mapbox://systemlevel.utah_units" }} />`

`<div>
        <Layer id={this.props.layerId} sourceId={this.props.sourceId} type="fill"
        layerOptions={this.props.unitLayerOptions}
[...]

Issue #261, opened 2017-06-12
GeoJSONLayer does not work with style-update


pcraciunoiu:
Initially in my code, I was getting this (doesn't happen in this example PR).

```
geojson-layer.js:77 Uncaught TypeError: Cannot read property 'setData' of undefined
    at GeoJSONLayer.componentWillReceiveProps (geojson-layer.js:77)
```

Fails right here:
```
    if (props.data !== data) {
      (map.getSource(id) as MapboxGL.GeoJSONSource).setData(props.data);
    }
```
`map.getSource(id)` is returning `undefined`.

I tried cloning the library and updating the example to show this:
https://github.com/alex3165/react-mapbox-gl/pull/260

In the example, there are no errors in the console, but the layer is not re-rendered. If you check out that branch and go to the "Style update" tab, you can see what I'm talking about.

If you switch the style the GEOJSONLayer is NOT re-rendered.

alex3165:
Hey, Thank you for the issue I will handle this with the `v2.0.0` milestone

alex3165:
https://github.com/alex3165/react-mapbox-gl/pull/269 should fix it

pcraciunoiu:
@alex3165 are there breaking changes if I upgrade to 2.0.0?

Should I wait to run off of that branch or is it stable?

alex3165:
I have released v2 under `@next` tag on npm, you can try out with `npm i react-mapbox-gl@next`, the API changed a bit, have a look at the new documentation: https://github.com/alex3165/react-mapbox-gl/blob/v2.0.0/docs/API.md

alex3165:
`2.0.3` is release as a main tag, GeoJsonLayer should now be updated correctly

pcraciunoiu:
@alex3165 this seems to work, but I am still getting the error when I switch styles (once each style change).
`Uncaught TypeError: Cannot read property 'setData' of undefined
    at GeoJSONLayer.componentWillReceiveProps (geojson-layer.js:93)`

Seems to happen here on the line with `getSource`:
```
    if (props.data !== data) {
      (map.getSource(id) as MapboxGL.GeoJSONSource).setData(props.data);
    }
```

Wondering if I am doing something odd, but it doesn't seem like it. The code for the layer looks like this:
```
<Map ...>
          <GeoJSONLayer
            data={generateDirectionArrow(...currentLocation, direction, zoom)}
            fillPaint={currentLocationDirectionStyles}
          />
</Map>
```

[...]

Issue #225, opened 2017-05-16
Changing Map style causes existing layers to disappear


emgeee:
I'm trying to allow someone to toggle between street view and a satellite view. I have several `<GeoJSONLayer>`'s mounted and working fine but when I change the `<Map>` style, those layers appear to disappear.

An error is thrown at line 114 of geojson-layer.ts saying `Uncaught TypeError: Cannot read property 'setData' of undefined`

z0d14c:
currently this is one of the big gotcha's with this lib. you need to remove the layers, change the style, then add the layers back for it to work properly. :(

emgeee:
Well that's unfortunate. Any idea why that is?

z0d14c:
nope, we have an existing bug in our app related to this issue, have fixed several times in our implementation over this lib but it keeps popping up; i dont mind doing the workaround (making sure all layers are removed, changing style, then adding them back), but it kinda sucks that it keeps popping up -- i think getting some guarantees and tests around how the map style changing behaves would be 👍 and if i can possibly find some time I'll try and work on it...

emgeee:
That's fair. I'm still trying to image with that workaround looks like since I imagine when you say "remove all layers, then add them back" you have to have a boolean somewhere that removes all inner components from the Map component, waits for a new `render()`, toggles the style, then toggle the boolean back so all the components are re-added. Did I get that right?

z0d14c:
yeah you got it right -- it looks like this feature is now working for me without the workaround, though. im on version 1.11.0

alex3165:
I will close this issue as it is been resolved with the version 2.0.0. Please use https://github.com/alex3165/react-mapbox-gl/issues/237 to be updated on the state of 2.0.0.

alex3165:
Released a fix from @SBRK with version `2.5.2`

PR #384, opened 2017-09-26
Geojsonlayer switch style fix

SBRK:
This is a fix for #320, #261, #225
The tricky part was that when the data of the `GeoJSONLayer` was updated, the data in `this.source` was not, and therefore when style changed and `initialize` was called, it used `this.source`, that had out of date data.

The other problem I encountered was mouse handlers being bound twice after a style change. I extracted the logic of removing source, layer, and unbinding handlers to call it both at style change and unmount.


src/geojson-layer.ts
```diff
@@ -237,6 +251,12 @@ export default class GeoJSONLayer extends React.Component<Props, {}> {
 
     if (props.data !== data) {
       (map.getSource(this.id) as MapboxGL.GeoJSONSource).setData(props.data);
+
+      this.source = {
```
alex3165: Good catch 👍

alex3165:
Thanks @SBRK , this is an amazing work on some of the core components of the library!

alex3165:
Released with version `2.5.2` 👍

PR #410, opened 2017-10-04
Move CSS injection into Map factory; FIXES #397

### LINKED ISSUE #397: Option to turn off CSS injection
At project we are currently working on, we are styling the Mapbox map via CSS, on multiple pages. On the pages without the React version, everything is fine, but on those with, this library injects the CSS and destroys the styling of the map to some degree.

I feel like there should be at least an option to turn off the CSS injection since that is somewhat intrusive.

hobofan:
CSS injection can now be deactivated by passing `injectCss: false` to the Map factory.

To prevent a breaking change, the legacy way of injecting CSS is still on by default and has to be deactivated via `window.MAPBOX_SKIP_CSS_INJECTION = true` before importing the package.

For the next release with breaking changes, the block in `index.ts` can simply be removed.


alex3165:
Thank you for the PR! I just have a question, could you clarify why it would be a breaking change to only inject the css in the map.tsx? Peoples will use the Map component anyway with this library so i think it should be alright to remove it from the index.tsx

hobofan:
> Peoples will use the Map component anyway with this library

Yup, with that assumption everything would be fine, and it definitely is not a crazy breaking change, but still one in the strictest sense.

In the constellation I've encountered the problem, it would be one. Mid-size code-base where mapbox-gl is once used via the React variant in this package and once without it. There, some styles would now disappear from the version without it.

I don't have any hard feelings either way, so if you think that it would be okay to remove it directly, I can alter the PR 🙂

alex3165:
@hobofan ok I see, yeah if you don't mind I am happy to take the risk and remove it from the index directly 👍 then we can merge this PR

hobofan:
@alex3165 Done!

PR #429, opened 2017-10-24
Update before in geojson layer

### LINKED ISSUE #127: Layer's before prop not updating
We need to compare prop.before so that it gets updated (map.moveLayer needs to be called) in componentWillReceiveProps

PR up #126 

relevant changes in `layer.ts` -- accidentally branched form another feature branch, so it may list changes from #122

oliverbienert:
This resolves #127 for geojson layers also, I hope.

### DIFF
--- a/src/geojson-layer.ts
+++ b/src/geojson-layer.ts
@@ -307,6 +307,12 @@ export default class GeoJSONLayer extends React.Component<Props, {}> {
 
     });
 
+    if (before !== props.before) {
+      ['symbol', 'line', 'fill', 'fill-extrusion', 'circle'].forEach(type => {
+        const layerId = this.buildLayerId(type);
+        map.moveLayer(layerId, props.before);
+      });
+    }
   }
 
   public render() {
--- a/src/geojson-layer.ts
+++ b/src/geojson-layer.ts
@@ -300,11 +300,13 @@ export default class GeoJSONLayer extends React.Component<Props, {}> {
           }
         }
       });
+
+      if (before !== props.before) {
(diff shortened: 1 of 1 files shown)


alex3165:
Hey @oliverbienert Thank you for the PR, I will have a look as soon as I have some free time 👍

src/geojson-layer.ts
```diff
@@ -300,7 +300,19 @@ export default class GeoJSONLayer extends React.Component<Props, {}> {
           }
         }
       });
+
+      if (before !== props.before) {
+        map.moveLayer(layerId, props.before);
+      }
+
     });
+
```
alex3165: you can remove from line 309 to 315, what you do line 304 to 306 is the same as it is included in a forEach which run for every layer types. See line 263. Once this is done I will merge this PR

oliverbienert: I'd rather remove line 304-306 which I added in a previous commit and forgot to remove later. Reason is, I can not be sure about order of keys that will come from `Object.keys(typeToLayerLUT)` as per definition this is an unordered collection of properties. So I listed the keys explicitly in line 311 to make sure I preserve order of geojson sublayers.
What do you think?

alex3165: I did some changes regarding the `typeToLayerLUT` object. Thanks for your help 👍 I kept your commits
cf: https://github.com/alex3165/react-mapbox-gl/pull/438

oliverbienert: Great! Nice solution.

alex3165:
Published with version `react-mapbox-gl@2.7.1`

PR #434, opened 2017-11-02
projected-layer: react to offset prop change

vehsakul:
I was unable to implement some functionality because projected layer reacts only to coordinates changes.


alex3165:
Hi @vehsakul, very sorry for the late response, I have added a check for the anchor prop as well.

cf: https://github.com/alex3165/react-mapbox-gl/pull/439/files

Thanks for your help 👍

alex3165:
Published with version `react-mapbox-gl@2.7.1`

vehsakul:
Thanks a lot!

PR #471, opened 2017-12-08
onSourceLoaded sometimes firing too soon

redbmk:
It looks like #453 wasn't quite complete. I found that sometimes on a slow network connection the source will be there but some properties like `bounds` don't exist yet. Fortunately there is a function `isSourceLoaded` to check for other things like outstanding network requests.


alex3165:
Thank you, that make sense

alex3165:
Released with tag `3.3.1`

PR #478, opened 2017-12-15
GeoJSON Layer: fix map crash

sergei-zelinsky:
Fix map crash when map style and GeoJSON data are changing simultaneously.


alex3165:
Hey @sergei-zelinsky , I am really sorry I missed it, I created a PR with the fix as well and a few other changes here https://github.com/alex3165/react-mapbox-gl/pull/481.

PR #480, opened 2017-12-19
Scale control is twice as width as it should be

kLaukkane:
_canvas.width seems to be twice as large as the actual clientWidth. In result, the scale control was being rendered twice as large as it should be; you can easily verify this using the example project and comparing the results to google map distance measurement, for example.

My fix uses clientWidth instead of canvas width, making the control the correct width. This leaves the control quite small, so I also adjusted the MIN_WIDTH_SCALE.


alex3165:
Hey @kLaukkane thanks for the PR, sorry for taking a long time to merge this, I wanted to spend some times testing it.

PR #502, opened 2018-01-22
Allow scroll events on HTML Layers

### LINKED ISSUE #217: Bug: Zoom in/out with scroll stops to work if mouse is over marker
PS library is super useful, because it works with CRA. Thanks

redbmk:
Fixes #217 
Could switch to `ReactDOM.createPortal` at some point if dropping support for React 15

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,8 +66,8 @@
   "peerDependencies": {
     "mapbox-gl": "0.44.0",
     "prop-types": "^15.5.10",
-    "react": "^15.6.1 || ^16.0.0",
-    "react-dom": "^15.6.1 || ^16.0.0"
+    "react": "^16.0.0",
+    "react-dom": "^16.0.0"
   },
   "devDependencies": {
     "@types/core-js": "0.9.43",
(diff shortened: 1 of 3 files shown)


alex3165:
@redbmk Could you rebase this PR from master?

redbmk:
@alex3165 Sure, just updated the PR

redbmk:
@alex3165 I found that the way I had it had too many issues with other elements now being behind the canvas and passing events through, such as popups and zoom controls.

I also experimented a bit with `ReactDOM.render` vs `ReactDOM.createPortal` and found that `createPortal` will append to the DOM element, whereas `render` will overwrite what's in there. Using `createPortal` keeps the code cleaner and allows us to just call that for each `Marker`.

If you want to create a `v4` branch I'll change the base of this PR to that. I've noticed a couple UI issues with some of the controls that I was thinking might require changing the API a bit, but now I'm thinking we could use `createPortal` instead (I'll make a separate issue or PR for this).

alex3165:
Hey @redbmk thanks for tackling this issue, I think it would make sense to have a proper solution with createPortal in a v4 branch. I created the [branch](https://github.com/alex3165/react-mapbox-gl/tree/v4) so feel free to open a PR against it.

redbmk:
Sounds good - I actually just changed this PR, which is now using `createPortal` to base off the `v4` branch

alex3165:
Thank you @redbmk, would you mind just rebasing on `v4` branch and I will get this merged

redbmk:
No problem - there ya go.

Issue #501, opened 2018-01-22
Update style causes error when using geoJsonSource


benrudolph:
I have a `Source` element that uses geoJsonSource:

```
<Source id="entities" geoJsonSource={{
                    data: entityGeoJson,
                    type: 'geojson',
 }} />
```
Whenever I update the `style` attritube on the `Map` element, it throws an error:

```
source.js:91 Uncaught TypeError: Cannot read property 'setData' of undefined
    at Source../node_modules/react-mapbox-gl/lib/source.js.Source.componentWillReceiveProps (source.js:91)
    at callComponentWillReceiveProps (react-dom.development.js:9795)
    at updateClassInstance (react-dom.development.js:9974)
    at updateClassComponent (react-dom.development.js:10224)
    at beginWork (react-dom.development.js:10605)
    at performUnitOfWork (react-dom.development.js:12573)
    at workLoop (react-dom.development.js:12682)
    at HTMLUnknownElement.callCallback (react-dom.development.js:1299)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:1338)
    at invokeGuardedCallback (react-dom.development.js:1195)
    at performWork (react-dom.development.js:12800)
    at batchedUpdates (react-dom.development.js:13244)
    at performFiberBatchedUpdates (react-dom.development.js:1646)
    at stackBatchedUpdates (react-dom.development.js:1637)
    at batchedUpdates (react-dom.development.js:1651)
    at Object.batchedUpdatesWithControlledComponents [as batchedUpdates] (react-dom.development.js:1664)
    at dispatchEvent (react-dom.development.js:1874)
```

This is with version 3.4.1.

Seems very similar to issues:
https://github.com/alex3165/react-mapbox-gl/issues/466
https://github.com/alex3165/react-mapbox-gl/issues/261

alex3165:
Released with tag `3.5.0` of [react-mapbox-gl](https://github.com/alex3165/react-mapbox-gl/releases/tag/v3.5.0)

PR #503, opened 2018-01-24
Checks to ensure that the source has been added to the map

benrudolph:
Addresses: https://github.com/alex3165/react-mapbox-gl/issues/501

This ensures that the map has the source before attempting to set the data. This is important because `componentWillReceiveProps` is fired before the style change handler. Since all the sources are removed after the style change, this throws an error.

### DIFF
--- a/src/source.ts
+++ b/src/source.ts
@@ -116,7 +116,8 @@ export default class Source extends React.Component<Props> {
     if (
       geoJsonSource &&
       props.geoJsonSource &&
-      props.geoJsonSource.data !== geoJsonSource.data
+      props.geoJsonSource.data !== geoJsonSource.data &&
+      map.getSource(this.id)
     ) {
       const source = map.getSource(this.id) as GeoJSONSource;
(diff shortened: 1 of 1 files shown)


src/source.ts
```diff
@@ -116,7 +116,8 @@ export default class Source extends React.Component<Props> {
     if (
       geoJsonSource &&
       props.geoJsonSource &&
-      props.geoJsonSource.data !== geoJsonSource.data
+      props.geoJsonSource.data !== geoJsonSource.data &&
```
benrudolph: What is the intended functionality for this line? Is this supposed to be a deep equal?

benrudolph:
@alex3165 wondering what you think of this? Currently this is a show stopper bug for me and I have to use my own fork in order to keep using this library

alex3165:
@benrudolph Thanks, it make sense, sorry for the delay, I will try to catch up on `react-mapbox-gl` as I will have some free time in march.

PR #509, opened 2018-01-30
Set the ZoomControl div to tabIndex -1 to remove from keyboard control flow

stephenmcm:
As Mapbox-gl provides keyboard interactions when the map canvas is focused having the zoom controls also keyboard accessible makes little sense. They are already annoying to get to as a user has to tab past the Mapbox logo, terms and conditions for Mapbox and OpenStreetMap *and* an Improve this map link just to trigger them.

Future improvements could be made by providing a pop-up or toast for keyboard users outlining the controls they have access to.

### DIFF
--- a/src/zoom-control.tsx
+++ b/src/zoom-control.tsx
@@ -132,6 +132,7 @@ export default class ZoomControl extends React.Component<Props, State> {
       <div
         className={className}
         style={{ ...containerStyle, ...positions[position!], ...style }}
+        tabIndex={-1}
       >
         <button
           type="button"


alex3165:
Hey, thanks for the suggestion, could you add a prop `tabIndex` to the `zoom-control` component instead of explicitly defining the value to -1. Also isn't tabIndex -1 the default value?

It would be nice to have tabIndex prop for `rotation-control` and `scale-control` components too.

alex3165:
Closing this pr in favour of https://github.com/alex3165/react-mapbox-gl/pull/525

PR #516, opened 2018-02-02
ZoomControl minus button stays hovered on mouseout

redbmk:
When minus is hovered, `this.state.hovered` equals `0`, so the state is not changed on `mouseout`. It works if you mouseout from the plus button, but not from minus.


alex3165:
Thanks @redbmk 👍

redbmk:
No problem, happy to help!

PR #518, opened 2018-02-05
Add border-bottom radiuses to compass rotation control

cyrilchapon:
Compass rotation control is the only child in his group, so adding bottom border radiuses.


alex3165:
Thanks for the PR 👍

PR #519, opened 2018-02-05
Add drag start event to Feature

alex3165:


cyrilchapon:
Other strategy in #521, fixing small bug in the process

PR #521, opened 2018-02-05
Feature events

cyrilchapon:
- Rework and refactor whole event logic
- Fix a bug making not-moved-click triggering onDrag[End/Start]
- Implement `onDrag` and `onDragStart` to add on `onDragEnd`

### DIFF
--- a/src/layer-events-hoc.tsx
+++ b/src/layer-events-hoc.tsx
@@ -185,15 +185,25 @@ function layerMouseTouchEvents(
 
       this.hover.map(id => {
         const child = children[id];
+
         const onDragStart = child && child.props.onDragStart;
         if (onDragStart) {
           onDragStart({ ...evt, map });
+        }
+
+        const onDrag = child && child.props.onDrag;
+        if (onDrag) {
+          onDrag({ ...evt, map });
         }
       });
     }
 
     // tslint:disable-next-line:no-any
     public onFeatureDrag = (evt: any) => {
+      if(!this.hasDragged) {
+        return;
+      }
+
       const children = this.getChildren();
       const { lngLat: {lng, lat} } = evt;
 
@@ -211,15 +221,16 @@ function layerMouseTouchEvents(
       this.forceUpdate();
 
 
-      // const { map } = this.context;
+      // Trigger onDrag
+      const { map } = this.context;
 
-      // this.hover.map(id => {
-      //   const child = children[id];
-      //   const onDrag = child && child.props.onDrag;
-      //   if (onDrag) {
-      //     onDrag({ ...evt, map });
-      //   }
-      // });
+      this.hover.map(id => {
+        const child = children[id];
+        const onDrag = child && child.props.onDrag;
+        if (onDrag) {
+          onDrag({ ...evt, map });
+        }
+      });
     }
 
     // tslint:disable-next-line:no-any
--- a/src/layer-events-hoc.tsx
+++ b/src/layer-events-hoc.tsx
@@ -198,7 +198,7 @@ function layerMouseTouchEvents(
 
     // tslint:disable-next-line:no-any
     public onFeatureDrag = (evt: any) => {
-      if(!this.hasDragged) {
+      if (!this.hasDragged) {
         return;
       }
 
@@ -234,7 +234,7 @@ function layerMouseTouchEvents(
     public onFeatureDragEnd = (evt: any) => {
       this.isGrabbing = false;
 
-      if(!this.hasDragged) {
+      if (!this.hasDragged) {
         return;
       }
 
--- a/src/layer-events-hoc.tsx
+++ b/src/layer-events-hoc.tsx
@@ -122,6 +122,7 @@ function layerMouseTouchEvents(
       map.dragPan.disable();
 
       this.isDragging = true;
+      map.once('mousemove', this.onDragDown);
       map.on('mousemove', this.onDragMove);
       map.once('mouseup', this.onDragUp.bind(this, 'mousemove'));
     };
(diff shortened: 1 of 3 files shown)


cyrilchapon:
This still requires unit testing though. But way more cleaner than my previous one, plus fixing a bug.

cyrilchapon:
Fix lints

src/layer-events-hoc.tsx
```diff
@@ -132,57 +137,118 @@ function layerMouseTouchEvents(
...
+      const endEvent = startEvent === 'mousedown' ? 'mouseup' : 'touchend';
+      const { map } = this.context;
+
+      this.isGrabbing = true;
+      this.hasDragged = false;
+
+      this.draggedChildren = [];
+      map.once(moveEvent, this.onFeatureDragStart);
+      map.on(moveEvent, this.onFeatureDrag);
+      map.once(endEvent, this.onFeatureDragEnd);
```
alex3165: Can you listen off those events `onFeatureDragEnd` / on `componentWillUnmount` ?

src/layer-events-hoc.tsx
```diff
@@ -132,57 +137,118 @@ function layerMouseTouchEvents(
...
+        // Set hover
         const { id } = feature.properties;
         this.hover = [id];
-        if (this.areFeaturesDraggable(children)) {
-          map.dragPan.disable();
-          this.isDragging = true;
 
-          map.on('touchmove', this.onDragMove);
-          map.once('touchend', this.onDragUp.bind(this, 'touchmove'));
+        if (!this.areFeaturesDraggable(children)) {
```
alex3165: This was an issue, instead of checking all the children feature `!this.areFeaturesDraggable(children)`, we should check for the current touched feature, something like: `!this.isFeatureDraggable(feature)`. Can we change this in the meantime?

alex3165:
Thanks a lot for the PR, good work 👍  I will test it asap

cyrilchapon:
Thanks bud.

I think we could release this.hover on touchend, on mobile shall we ?

I didn't dare to fix it in thé process because of side effects

cyrilchapon:
I didn't get what you expected as an update, concerning event off handling ?

cyrilchapon:
@alex3165, can we expand on the 2 updates you asked for ?
Maybe slack, gitter or something ?

alex3165:
For `areFeaturesDraggable `, you could replace line `143` with: 
```
        if (!feature.properties.draggable) {
          return;
        }
```
and in `layer.ts` line `131`:
```
    properties: { ...props.properties, draggable: props.draggable, id }
```

alex3165:
For the event listeners, I didn't see you are listening off `map.on(moveEvent, this.onFeatureDrag);` already here:
```
      map.once(endEvent, () => {
        map.off(moveEvent, this.onFeatureDrag);
        this.draggedChildren = undefined;
      });
```

the other ones are just `.once` listeners so I guess they automatically unbind the event listener after the first event being triggered so it should be alright

cyrilchapon:
Ok I'll fix today.

alex3165:
Thanks for the work, I have tested it, it works fine, once you do the changes for `areFeaturesDraggable ` we can get this merged. It would be nice to add an issue about tests for this bit which has heavy logic / states.

cyrilchapon:
@alex3165, trying to fixing it, I realized it might need some deeper rework, mostly :
- Should it drag **every** draggable features under mouse ? Or just last / first ?
- "Global" logic (dragPan mostly) should be refactored and done only once (this is in the loop currently)
- Why is there a loop in touchStart and not in mouseDown ?

alex3165:
Ah yeah I didn't realised it is inconsistent between `touchStart` and `mouseDown`, so after a deeper look we don't rely on `evt.features` in `onMouseDown` because we already push the feature ids in `this.hover` in `onMouseEnter`. So in the end we drag everything that is in `this.hover`. Can you just move out of the loop everything except the bit of logic that add the feature ids to `this.hover` if the feature is draggable and do the same logic (check for the feature being draggable before pushing to `this.hover`) in `onMouseEnter` ?

cyrilchapon:
Ok. So basically :
- Touch strategy : onTouchStart sets this.hover, onTouchEnd empty it
- Mouse strategy: mouseOver sets this.hover, mouseLeave empty it

And I drag this.hover

alex3165:
Yes exactly 👍

cyrilchapon:
Done as you suggested but there is still an inconsistency : If two features are stacked, one draggable and not the other, it would drag both of them I guess

I think there should be a rework on this whole file, accepting only one `hover` at a time - say the top most draggable - and drag features one by one

alex3165:
@cyrilchapon it would drag both of them because we should be checking if each single feature is draggable or not when iterating on `evt.features` instead of checking outside with `this.areFeaturesDraggable(children)` I will branch off this PR and open a new one

cyrilchapon:
@alex3165 yes.

I thinking we could choose between :
- Keeping only one, draggable (highest "z-index"), feature inside `this.hover`
- Or keeping everything under mouse inside `this.hover` and dragging only the highest z-index one in event handlers

alex3165:
True I moved every iteration to use `this.hover` as soon as possible and only `evt.features` to fill `this.hover` and removed `this.getChildren()[feature.id]` then the later one is only used to get the children element from the id in `this.hover` to the element in `this.getChildren()`.

You can check the last commit: https://github.com/alex3165/react-mapbox-gl/pull/528

alex3165:
Closing this PR in favour of https://github.com/alex3165/react-mapbox-gl/pull/528 that has been merged

PR #500, opened 2018-01-21
Upgrade mapbox-gl to 0.43.0

jthetzel:
1) Upgrade mapbox-gl to 0.43.0
2) Remove layers prior to removing source fixes #493 
3) Upgrade react, react-dom in example fixes #499 
4) Added toggleLayer demo to demonstrate #493 is now fixed

Issue #493, opened 2018-01-08
Error removing layer when using mapbox-gl 0.43.0


jthetzel:
Mapbox-gl v0.43.0 introduces a breaking change that throws an error if a source is removed while still in use (e.g. remove a source prior to removing layers that use the source, https://github.com/mapbox/mapbox-gl-js/pull/5562 ). This was discovered in #491.

Example error:
```
mapbox-gl.js:546 Error: Source "example_id" cannot be removed while layer "example_id_marker" is using it.
    at t.removeSource (mapbox-gl.js:406)
    at e.removeSource (mapbox-gl.js:520)
    at Source.componentWillUnmount (source.js:71)
    at callComponentWillUnmountWithTimerInDev (react-dom.development.js:11117)
    at HTMLUnknownElement.callCallback (react-dom.development.js:1299)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:1338)
    at invokeGuardedCallback (react-dom.development.js:1195)
    at safelyCallComponentWillUnmount (react-dom.development.js:11125)
    at commitUnmount (react-dom.development.js:11415)
    at unmountHostComponents (react-dom.development.js:11356)
```

jthetzel:
For reproducible example in example/demos, see toggle-layer demo at: https://github.com/jthetzel/react-mapbox-gl/commit/4872734cc807cdfa0ff11ecd2b8b2631b3f5efd8

Wykks:
Yup I had to deal with this change in ngx-mapbox-gl too (https://github.com/Wykks/ngx-mapbox-gl/issues/10).
@alex3165 Instead of calling mapbox-gl method directly, we can buffer the removes request, then in the same frame if possible process the removes in a defined order.

alex3165:
Released with tag `3.5.0` of [react-mapbox-gl](https://github.com/alex3165/react-mapbox-gl/releases/tag/v3.5.0)

redbmk:
Odd - I'm still getting this error when using mapbox-gl `0.43.0` and react-mapbox-gl `3.5.0`. Same with `0.44.0` and `3.5.1`. I might be doing something a little differently than the examples. I'll see if I can tweak the them to reproduce this.

redbmk:
It turns out I was seeing the bug when swapping out tileJSON sources. I opened PR #530 to fix it.

alex3165:
Published a fix from @redbmk in version `3.8.0`, enjoy! 👍

PR #530, opened 2018-02-09
Fix error updating tileJsonSource in mapbox-gl^0.43

redbmk:
This is the same error as #493. The bug was fixed in #500 for when a source is removed, but if a `tileJsonSource` is changed, the source is removed from the map and added again. This modifies the fix from #500 to remove the relevant layers, but then adds them back in after re-adding the source.

### DIFF
--- a/src/source.ts
+++ b/src/source.ts
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
-import { Map, GeoJSONSource, GeoJSONSourceRaw } from 'mapbox-gl';
+import { Map, GeoJSONSource, GeoJSONSourceRaw, Layer } from 'mapbox-gl';
 import { SourceOptionData, TilesJson } from './util/types';
 
 export interface Context {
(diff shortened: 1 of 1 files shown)


src/source.ts
```diff
@@ -79,6 +83,29 @@ export default class Source extends React.Component<Props> {
...
+          const { id: before } = layers[idx + 1] || { id: undefined };
+          return { ...layer, before };
+        })
+        .filter(layer => layer.source === this.id);
+
+      layers.forEach(layer => map.removeLayer(layer.id));
+
+      map.removeSource(this.id);
+
+      return layers.reverse();
```
alex3165: Why do you need to reverse the layers?

alex3165:
Hey @redbmk Thanks for the PR 👍 I just left 1 comment

src/source.ts
redbmk: This is because `map.addLayer` has an option for `before`, but not `after` ([`addLayer(layer, before?)`](https://www.mapbox.com/mapbox-gl-js/api#map#addlayer)). Say for example you have layer IDs that look like `[ 'asdf', 'zxcv', 'qwer' ]`, then we generate something like `[ { id: 'asdf', before: 'zxcv' }, { id: 'zxcv', before: 'qwer' }, { id: 'qwer', before: undefined } ]`.

If you try to add them back in the same order, `asdf` can't go before `zxcv` because it's not in there yet. So reversing it makes it so that `qwer` goes in at the end, then `zxcv` gets added in before `qwer`, etc.

I would love for the code to be more clear so this doesn't get changed later. Maybe we could add a comment, or can you think of a way to rewrite it that it's more obvious?

alex3165:
Published with tag `3.8.0` 👍 thanks and enjoy!

PR #553, opened 2018-03-08
Fix issue where un-changed source will be re-added

Rob-ot:
Mapbox-gl allows `tiles` to be an array instead of a string in order to rotate between multiple url endpoints. (for parallelizing requests)

We were running into an issue where the source would be re-added causing all tiles to be reloaded and flicker.

`deep-equal` works fine on strings and array so this is a simple fix.

Thanks.

### DIFF
--- a/src/source.ts
+++ b/src/source.ts
@@ -1,5 +1,7 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
+const isEqual = require('deep-equal'); //tslint:disable-line
+
 import { Map, GeoJSONSource, GeoJSONSourceRaw, Layer } from 'mapbox-gl';
 import { SourceOptionData, TilesJson } from './util/types';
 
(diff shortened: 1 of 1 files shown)


alex3165:
Hey @Rob-ot sorry for the late reply, I would not consider this as a bug. Using reference equality to check for an update is a design choice. If you don't want to update the `Source` when changing `tileJsonSource.tiles`  you should make sure not to break the reference equality of the tiles object.

The problem with `isEqual` is that it is slower than `===`.

Rob-ot:
@stdmn Sure. All I did was make the tiles array a variable that is reused, so:

at the top of the file `const streetMapTiles = ["link_to_file/{z}/{x}/{y}.pbf"]`

then use those tiles:
```
{
    'type': 'vector',
    'tiles': streetMapTiles,
    "minzoom": 8,
    "maxzoom": 14,
}
```

Hope that helps!

stdmn:
Thank you! That fixed it.

PR #558, opened 2018-03-13
add transformRequest map option

m-abboud:


alex3165:
Thanks for the PR @m-abboud 👍

cyrilchapon:
With a little piece of documentation ? 😃

DanielUche:
node_modules/react-mapbox-gl/lib/map.d.ts
(85,13): Subsequent property declarations must have the same type.  Property 'transformRequest' must be of type 'Function | undefined', but here has type 'RequestTransformFunction | undefined'.



the above error occurs when using this package with typescript

PR #559, opened 2018-03-13
make map interaction handlers like doubleClickZoom toggle-able through props

m-abboud:
It's useful to be able to disable things like doubleClickZoom when switching to a drawing tool.

### DIFF
--- a/src/map.tsx
+++ b/src/map.tsx
@@ -58,6 +58,13 @@ export interface Props {
   animationOptions?: Partial<AnimationOptions>;
   flyToOptions?: Partial<FlyToOptions>;
   children?: JSX.Element | JSX.Element[] | Array<JSX.Element | undefined>;
+  doubleClickZoom?: boolean;
+  touchZoomRotate?: boolean;
+  dragPan?: boolean;
+  dragRotate?: boolean;
+  scrollZoom?: boolean;
+  keyboard?: boolean;
+  boxZoom?: boolean;
 }
 
 export interface State {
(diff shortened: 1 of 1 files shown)


src/map.tsx
```diff
@@ -363,9 +370,28 @@ const ReactMapboxFactory = ({
         map.setStyle(nextProps.style);
       }
 
+      const handlerNames = ['doubleClickZoom', 'dragPan', 'dragRotate',
+        'touchZoomRotate', 'scrollZoom', 'boxZoom', 'keyboard']
+      for (const name of handlerNames) {
```
alex3165: Would you mind changing to forEach instead, so that it is consistent with the rest of the codebase?

alex3165:
Good job but it would be nice to move the logic to a function in `map-events.tsx`, move `handlerNames` outside of the function and use that function in componentWillReceiveProps of the map? Same as what have been done for `updateEvents`

alex3165:
Actually could you explain why you need to do that? is it not what `updateEvents` already do? if you don't provide the event handler then `updateEvents` function will listen off the event and won't attach any new handler, maybe I am missing something

m-abboud:
dragPan and related handlers are the ones built into mapbox not user provided event handlers. updateEvents appears to be for passing in your on onDrag event handler function vs dragPan which is you just need to tell mapbox to turn it off or on?

You could implement your own dragPan and pass it in with the current user event handler props to gain the ability to toggle it on and off but that's additional work someone using this lib shouldn't have to do.

alex3165:
Closing this for inactivity, feel free to reopen if needed

PR #560, opened 2018-03-13
allow manually setting a features geometry type

m-abboud:
This is needed because the auto geometry type selection does not work when you want to use both a line style and a fill style for a single polygon's coordinates


m-abboud:
To add some clarity, passing in a polygon's coordinates  in format like [[[0,0]...] to a feature with a line type style will cause an error because a line type style will assume the geometry is in a format with no wrapping array like [0,0],[1,1]. (I think that was the use case at least been a minute since I added this)

Line type styles are necessary for a polygon's outline/border so you have to use a seperate fill style and a line style if you want a filled polygon with a border.

PR #565, opened 2018-03-29
Correct API.md

jezhiggins:
Center co-ordinates should, in defiance of the original documentation, be in latitude, longitude order.  The example coords here and else where in the documentation are clearly lat, long and, more to the point, that's what you need to do to get your map to display properly.

From map.tsx 
  const defaultCenter = [-0.2416815, 51.5285582];
which I assume means somewhere in east London rather than in the middle of the sea off the coast of Somalia.  This is clearly lat, long.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -48,7 +48,7 @@ const Map = ReactMapboxGl({
 ### Component Properties
 - **style** *(required)* : `string | object`  Mapbox map style, if changed, the style will be updated using `setStyle`.
 - **center** *(Default: `[ -0.2416815, 51.5285582 ]`)*: `[number, number]` Center the map at the position at initialisation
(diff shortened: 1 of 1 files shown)


mklopets:
The documentation is correct, `-0.2416815, 51.5285582` is longitude, latitude (i.e in GeoJSON Position format) for a location in London. Seems to me that you're confusing what latitude/longitude mean (tbh, I confuse them all the time myself).

For Mapbox's docs, find `options.center` under https://www.mapbox.com/mapbox-gl-js/api#map.

Google Maps uses the lat, lng format. Plop this into the query parameter and you get this https://www.google.com/maps/?q=-0.2416815,%2051.5285582, i.e the sea you described.

PR #573, opened 2018-04-12
Add offset option to zoomOnClick when clicking on Cluster

nicholasbraun:


nicholasbraun:
I'll close this since I wan't to put the logic for padding and offset in a seperate zoomOnClickOptions prop.

PR #584, opened 2018-05-01
Expose esm build

tusbar:
Not sure this is the right way to do it, not really familiar with typescript.
Let me know if anything needs to change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 
 vendor/
 lib/
+lib-esm/
 dist/
 es/
 www/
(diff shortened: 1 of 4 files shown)


brendanmoore:
To keep the `tsconfigs` consistent you can extend and only change this parts you need, e.g: in `tsconfig-esm.json`:

```json
{
  "extends": "./tsconfig",
  "compilerOptions": {
    "module": "esnext",
    "outDir": "lib-esm",
  }
}
```

alex3165:
Hey, @tusbar thanks for the update, I would be glad to merge this if you could update `tsconfig-esm.json` to extend `tsconfig.json` as mentioned by @brendanmoore

tusbar:
@alex3165 done! sorry for the delay.

PR #611, opened 2018-08-17
Type for handling subset of the Anchor keys at offset props

zoltan-mihalyi:
The documentation of Popup says:

"When object is passed, it must contain keys for different anchor positions and values as the offset (number or number[])"
But the offset property type is the following:

```typescript
  offset?: number | number[] | Point;
```
We added the object type: OptionalOffset

### DIFF
--- a/src/popup.tsx
+++ b/src/popup.tsx
@@ -3,12 +3,12 @@ import ProjectedLayer from './projected-layer';
 import * as GeoJSON from 'geojson';
 import { getClassName } from './util/classname';
 import { Point } from 'mapbox-gl';
-import { Anchor } from './util/types';
+import { Anchor, OptionalOffset } from './util/types';
 
 export interface Props {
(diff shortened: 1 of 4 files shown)


mklopets:
I don't see anything in the MapboxGL docs about them allowing specifying offsets for subsets of the anchor positions. Correct me if I'm missing something!

If passing subsets isn't allowed, what's wrong with the current `AnchorsOffset` interface that is currently in the `offset` property's union type?

zoltan-mihalyi:
In overlays.ts there is a line, in which the defaultPoint is used, when the given offset is not set:

```typescript
res[anchor] = Point.convert(offset[anchor] || defaultPoint);
```

The feature of using subsets of offset is alredy there, OptionalOffset reflects that.

zoltan-mihalyi:
Can you accept this if I rebase this onto master?

PR #612, opened 2018-08-17
preferredAnchor property

zoltan-mihalyi:
Currently there is no way to implement a good context menu on the map using `Popup` or `ProjectedLayer`, because if you set the `anchor` property to top-left, the popup uses that anchor even if the bottom of the popup is outside of the map.
We tried to solve this issue using an optional `preferredAnchor` boolean, which causes the anchor flip vertically or horizontally if needed

### DIFF
--- a/src/popup.tsx
+++ b/src/popup.tsx
@@ -18,6 +18,7 @@ export interface Props {
   style?: React.CSSProperties;
   className?: string;
   tabIndex?: number;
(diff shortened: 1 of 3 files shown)


alex3165:
Hey @zoltan-mihalyi Could you rebase from master and add the property to the documentation ?

zoltan-mihalyi:
Done

Issue #578, opened 2018-04-23
Cannot build with yarn


hyperknot:
I'm trying to run `yarn build` and `yarn start` in the examples folder but none of them works.

1. Git clone master (or checkout 3.8.0 tag)
2. `yarn`, OK
3. `yarn build` terminates in: 

```
Test Suites: 12 passed, 12 total
Tests:       48 passed, 48 total
Snapshots:   0 total
Time:        4.291s
Ran all test suites.


node_modules/@types/enzyme/node_modules/@types/react/index.d.ts(2250,13): error TS2403: Subsequent variable declarations must have the same type.  Variable 'a' must be of type 'DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>', but here has type 'DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>'.
node_modules/@types/enzyme/node_modules/@types/react/index.d.ts(2251,13): error TS2403: Subsequent variable declarations must have the same type.  Variable 'abbr' must be of type 'DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>', but here has type 'DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>'.
```

And about a hundred or so occurrences afterwards. Ending with 

```
  Named property 'ref' of types 'ClassAttributes<T>' and 'ClassAttributes<T>' are not identical.
src/cluster.tsx(129,21): error TS2345: Argument of type '{ clusterPoints: Cluster[]; }' is not assignable to parameter of type 'State | ((prevState: Readonly<State>, props: Props) => State | Pick<State, "clusterPoints">) | Pi...'.
  Type '{ clusterPoints: Cluster[]; }' is not assignable to type 'Pick<State, "clusterPoints">'.
    Types of property 'clusterPoints' are incompatible.
      Type 'Cluster[]' is not assignable to type 'Feature<Point>[]'.
        Type 'Cluster' is not assignable to type 'Feature<Point>'.
          Types of property 'id' are incompatible.
            Type 'string | number | undefined' is not assignable to type 'string | undefined'.
              Type 'number' is not assignable to type 'string | undefined'.
src/cluster.tsx(175,49): error TS2345: Argument of type 'Cluster[]' is not assignable to parameter of type 'Feature<any>[]'.
  Type 'Cluster' is not assignable to type 'Feature<any>'.
[...]

PR #618, opened 2018-08-25
React.createContext & typescript 2.9.1

### LINKED ISSUE #520: Upgrade to Typescript 2.7

brendanmoore:
This PR updates the map context to use the new `React.createContext` pattern. This has also had the bonus of removing `prop-types` as a dependency. In addition typescript is upgraded to 2.9.1. (closes #520, #578)

A cool thing we could do is expose `MapConsumer` so users could have direct access to the map instance in context without having to rely on an `onStyleLoaded` side-effect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,8 +59,8 @@ const Map = ReactMapboxGl({
 
 // in render()
 <Map
-  style="mapbox://styles/mapbox/streets-v9"
-  containerStyle={{
+  mapStyle="mapbox://styles/mapbox/streets-v9"
+  style={{
     height: "100vh",
     width: "100vw"
   }}>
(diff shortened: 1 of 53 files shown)


brendanmoore:
So it seems that the fixes in #502 has created an issue with the cluster markers no longer being rendered as a child of the expected `parentElement` div use in `Cluster.zoomToClusterBounds`. If we want to keep the API the same i.e. having access to the `parentElement` for `<Cluster style={...} className={...} />` the `ProjectLayer` API will probably need to be refactored to handle optional containers for the `ReactDOM.createPortal`

brendanmoore:
5e8205a Has resolved the `Cluster.zoomToClusterBounds` issue by hoisting the `createPortal` implementation in #502 to `map.tsx`

brendanmoore:
@mklopets I rebased masters updates into my `v4` fork branch mainly to get the updates to GeoJSON and the latest mapbox-gl version, it has obviously left things a bit messy. It might make sense for the someone with write access to merge master into `v4` to clean it up a bit? What do you think?

mklopets:
Sure, I'll rebase the v4 branch on top of master. Thanks for your work btw, @brendanmoore!

brendanmoore:
I am going to clean up the commits close this PR and resubmit.

brendanmoore:
Replaced by #619

PR #619, opened 2018-08-29
React.createContext, TypeScript 2.9.1 & more

brendanmoore:
- Refactor deprecated old React context api to new pattern,
 - Update TypeScript version to 2.9.1.
 - Hoist `React.createPortal` to `map.tsx` fixes `Cluster.zoomToClusterBounds`
 - Remove `prop-types` and `recompose` (no longer required)

supersedes #618
closes #520  
closes #578

PR #620, opened 2018-08-30
Fix demos regression

brendanmoore:
Ref: #619 

@mklopets Quick fixes to get the demos working again...

This:
https://github.com/alex3165/react-mapbox-gl/compare/v4...brendanmoore:feature/react-create-context-typescript-update?expand=1#diff-3003f9bd42e62ece8399351cef1f10c7R130
is interesting, I couldn't actually figure out why the call signature types were incompatible so had to resort to `any` to get it working. I think I am missing something.


example/tsconfig.json
```diff
@@ -13,13 +14,11 @@
     "noUnusedLocals": true,
     "outDir": "./dist",
     "removeComments": true,
+    "skipLibCheck": true,
     "sourceMap": true,
-    "strictNullChecks": true,
+    "strictNullChecks": false, // Prevents: JSX element type 'ReactElement<any> | null' is not a constructor function for JSX elements. Type 'null' is not assignable to type 'ElementClass'
```
mklopets: would be nice to keep `strictNullChecks` on... I don't really understand why this error is there in the first place tbh

example/package.json
```diff
@@ -34,7 +34,7 @@
     "html-webpack-plugin": "2.30.1",
     "raw-loader": "0.5.1",
     "ts-loader": "2.3.7",
-    "typescript": "2.5.3",
+    "typescript": "2.9.2",
```
mklopets: It's `^3.0.3` on the actual package (I changed it), please make this match

example/src/demos/htmlCluster.tsx
```diff
@@ -127,7 +127,7 @@ class HtmlCluster extends React.Component<Props, State> {
         onMove={this.onMove}
         style={mapContainerStyle}
       >
-        <Cluster ClusterMarkerFactory={this.clusterMarker}>
+        <Cluster ClusterMarkerFactory={this.clusterMarker as any}>
```
mklopets: Try this (`git apply dontcast2any.patch`): https://gist.github.com/mklopets/461f206582af8dca0e96ce7a29c56f59

mklopets: still requires a cast but a much more reasonable one than to `any`

example/tsconfig.json
brendanmoore: It is because of this: https://github.com/Microsoft/TypeScript/issues/11955

brendanmoore:
Thanks for the `ClusterMarkerFactory as any` pointers... It did occur to me that there is some (for a least) easy to mis-read types around the codebase i.e. `Array<React.Component<MarkerProps>>` and `Array<React.ReactElement<MarkerProps>>` i wonder if it would help to export:

```typescript
export type MarkerElement = React.ReactElement<MarkerProps>;
export type MarkerComponent = React.Component<MarkerProps>;
```
Not sure if this is a good idea?

example/tsconfig.json
mklopets: That issue was supposedly fixed though(?)

mklopets:
Will look at the other comment tomorrow - not at a computer right now and don't know the code there by heart

brendanmoore:
@mklopets Looks like an `@types/react` update fixed the `strictNullChecks` issue, _should_ be good to go 😬

PR #623, opened 2018-09-01
Layer Event Handlers (Rough outline)

stdmn:
Not sure if this will work. I have the old JS version but struggling to get this working as a TS version. This is the basic idea of what needs to be done but will probably have to be refined.


daniel-hauser:
Resolved in #675

PR #640, opened 2018-10-03
Don't pass window to injectCSS

tstirrat15:
## Motivation
This is a low-hanging fruit in terms of making this library easier to use in a server context.

The `injectCSS` function references `window`, which is either a global variable in the frontend context or is undefined in the backend context. It makes more sense for injectCSS to reference it directly, as it's already making checks to see whether `window` conforms to the expected shape.

## Changes
* Remove window pass from map.tsx
* Add a check to ensure that window is not undefined before continuing to manipulate `window`

### DIFF
--- a/src/map.tsx
+++ b/src/map.tsx
@@ -144,7 +144,7 @@ const ReactMapboxFactory = ({
   transformRequest
 }: FactoryParameters) => {
   if (injectCss) {
-    injectCSS(window);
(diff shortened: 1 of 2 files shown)


alex3165:
Hey @tstirrat15 thanks for the work. Although v4 is no more injecting a custom style but it is requiring the css from mapboxgl-js directly

PR #651, opened 2018-10-29
Fix layer rendering

sheerun:
This PR fixes rendering of layers with geoJSONSourceOptions. The problem was two-fold:

- data has been overridden by default values, not the other way around (fixed test)
- generated sourceId was the same as layer id which apparently prevented it from being rendered

I used convention of appending `-source` to generated layer but I'm open for suggestions.


mikhail-angelov:
fix is working

Issue #654, opened 2018-11-08
Export MapContext to support custom controls 


daniel-hauser:
Using the new context API added in #619, the only way for a custom component to get the mapbox map instance is to get the `Map` ref, then extract the map from the state.

Using `MapContext.Consumer` (or alternatively, `withMap`) will make the code of custom controllers better.

amaurymartiny:
I believe this can be closed?

mklopets:
Released this in [v4.4.0](https://github.com/alex3165/react-mapbox-gl/releases/tag/v4.4.0)

PR #533, opened 2018-02-12
rename style to mapStyle, containerStyle to style

redbmk:
Fixes #498

### Component Properties

- **style** &rightarrow; **mapStyle**
  - **style** has been renamed to **mapStyle** to avoid confusion with DOM styles.
- **containerStyle** &rightarrow; **style**
  - **containerStyle** has been renamed to **style** to adhere to standard React convention.
  - The **style** prop now controls the DOM styles for the container element.

PR #679, opened 2019-01-08
Fix Marker & Cluster mouse events

### LINKED ISSUE #217: Bug: Zoom in/out with scroll stops to work if mouse is over marker

### LINKED ISSUE #562: Feature: Allow for clicking on marker but pass-through for drag
My use case I'm unable to implement:

My map has multiple markers that user is able to click for popup, but there are so many of them so if he/she tries to drag the map, the finger or cursor usually lads on one of markers. Ideally the click event on marker would cause popup to show, but drag event would pass-through to underlying map.

Is it possible with current implementation? I think this could be solved by catching drag events on marker and forwarding to map? But currently marker doesn't support onDragStart.

brendanmoore:
This fix was original merged (via #619) into an early v4 branch which has been abandoned. Using `createPortal` fixes capturing move/scroll/zoom events on Marker & Cluster elements

Closes #217 (again)
Closes #562

### DIFF
--- a/src/map.tsx
+++ b/src/map.tsx
@@ -8,6 +8,7 @@ import {
   updateEvents
 } from './map-events';
 import { MapContext } from './context';
+import { createPortal } from 'react-dom';
 const isEqual = require('deep-equal'); //tslint:disable-line
 
 export interface PaddingOptions {
@@ -374,15 +375,19 @@ const ReactMapboxFactory = ({
 
     public render() {
       const { containerStyle, className, children } = this.props;
-      const { ready } = this.state;
+      const { ready, map } = this.state;
+      const container =
+        ready && map && typeof map.getCanvasContainer === 'function'
+          ? map.getCanvasContainer()
+          : undefined;
       return (
-        <MapContext.Provider value={this.state.map}>
+        <MapContext.Provider value={map}>
           <div
             ref={this.setRef}
             className={className}
             style={{ ...containerStyle }}
           >
-            {ready && children}
+            {ready && container && createPortal(children, container)}
           </div>
         </MapContext.Provider>
       );


janhartmann:
I still have an issue where I can not mouse wheel over markers. Pointer events none is not an option for me, as I need onClick. 

I am using 4.2.0 and the demo here has the same issue: http://alex3165.github.io/react-mapbox-gl/demos

brendanmoore:
@janhartmann This PR should fix your issue

janhartmann:
Ah okay, I had a hard time figuring out all these issues. I hope to see it merged in asap :-)

redbmk:
@alex3165 @brendanmoore any movement on this? I just pulled the latest react-mapbox-gl only to realize this hadn't been merged yet. Just found another v4 PR I had made got lost as well (#533) 😞

redbmk:
I found a workaround for now...basically just manually implementing the `createPortal` logic in your own component.

```
import { withMap } from 'react-mapbox-gl/lib-esm/context';

const MyLayer = withMap(({ map }) => createPortal(
  <Popup>Hello World</Popup>
  map.getCanvasContainer(),
));
```

Eventually `withMap` will likely be exported directly from `react-mapbox-gl` (see #654)

rob137:
Would love to see this merged.  Is there anything I could do to help?

mklopets:
Thanks @brendanmoore!

mklopets:
Released in `v4.3.0`: https://github.com/alex3165/react-mapbox-gl/releases/tag/v4.3.0

mklopets:
Hmm, this seems to have caused https://github.com/alex3165/react-mapbox-gl/issues/725

mklopets:
Reverted the default functionality to not use a portal in [v4.5.0](https://github.com/alex3165/react-mapbox-gl/releases/tag/v4.5.0), see the release & issue #725 for more.

In hindsight, this was a breaking change and should have either gone into `v4.0.0` or waited for the next major release.

sgup:
Is the marker fix still working? Its still capturing scroll events on `v4.8.1` for me.

redbmk:
Unfortunately if you need some components to render in the portal and others not to (e.g. you want popups to capture events but markers to pass through), you'll still need to use your own workaround to render just the components you want in a portal. Instead of adding `renderChildrenInPortal`, you would need to do something like this:

```
import * as React from 'react';
import { MapContext } from 'react-mapbox-gl';

export const PassThroughMarker = props => {
  const map = React.useContext(MapContext);
  const container = map.getCanvasContainer();

  return container && createPortal(<Marker {...props} />, container);
};
```

PR #693, opened 2019-02-10
Add touch events to Layer/Feature/GeoJSONLayer

stdmn:
Adds onTouchStart and onTouchEnd to Layer, Feature and GeoJSONLayer for touch devices


sdll:
Thank you for the PR! Could it also cover touch events in the onClick handler, or `onTouchEnd` is enough in your opinion? 

Worth mentioning that the mapbox native event handler [deals](https://docs.mapbox.com/mapbox-gl-js/example/popup-on-click/) with touch clicks by default.

stdmn:
@sdll Do you mean that onClick should by default also include touch events or is the question whether or not we should also include the other touch events (onTouchMove, onTouchCancel, etc)?

The PR includes onTouchEnd and onTouchStart. For my purposes this is sufficient but not sure if others need onTouchMove or onTouchCancel.

sdll:
@stdmn, sorry for the confusion, turns out that onClick works on mobile as well, unless mapbox-gl-draw is added ([this](https://github.com/mapbox/mapbox-gl-draw/pull/869) should solve it). 

Thanks for your work!

michelebanfi:
Does anyone have an update on the PR?

Gwillison415:
@sdll We have a use case for the right mouse button click `onContextMenu` that looks almost exactly like this PR for `onTouchCancel`. I understand your comments above about these particular events - but how would you feel about a PR for that feature coming your way?

gianlu01:
> 
> 
> Does anyone have an update on the PR?
 
we need it

PR #711, opened 2019-04-03
Mark fitBounds events triggered by this library

sheerun:
This is very useful for `onDragEnd`, `onZoomEnd` and `onBoxZoomEnd` etc. event handlers to detect reason of drag/zoom (it can be either because user dragged map / clicked zoom button, or because fitBounds has been changed).

With this change each zoom/drag event triggered by this lib is marked with `{ auto: true }` which can be used for example in followng way:

```es6
            onBoxZoomEnd={(_, event) => {
              if (event.auto !== true) {
                this.performSearch()
              }
            }}
```


alex3165:
Hey @sheerun could we add this to the doc? I would also rename the property from `auto: boolean` to something like `fitboundUpdate: boolean`

sheerun:
Done

alex3165:
Thank you for you work, I will release asap

PR #712, opened 2019-04-03
Render markers into proper container

sheerun:
Currently markers are rendered outside of mapbox container which can cause lots of issues with proper styling of markers and managing their z-index. With this fix they are rendered inside canvas container and so it's much easier to keep them rendering within it and freely set map position, like `position: fixed`


mklopets:
I believe this should be addressed by #679

PR #715, opened 2019-04-09
Fix the lngLat reversal in the routes.json example file

nihalgonsalves:
`routes.json` has `lat` and `lng` reversed, and the `allShapes.ts` example that uses them also mapped them in the wrong order (`lat, lng` instead of `lng, lat`). It thus looked correct and the points were correctly placed in London visually - but the data is incorrect. This PR fixes the data and the mapping.


mklopets:
Nice catch, thanks!

PR #716, opened 2019-04-11
Upgrade react and fix `withMap()` parameter type

gesi:
This makes it possible to use function components with react@latest (16.8.13 at
time of commit) without TypeScript errors.

With `React.ComponentClass`, TypeScript produces this error when using a `FunctionComponent`

```
error TS2345: Argument of type '<T>(props: T) => Element' is not assignable to parameter of type 'ComponentClass<any, any>'.
  Type '<T>(props: T) => Element' provides no match for the signature 'new (props: any, context?: any): Component<any, any, any>'.
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -173,9 +173,9 @@
       "dev": true
     },
     "@types/prop-types": {
-      "version": "15.5.6",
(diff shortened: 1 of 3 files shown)


gesi:
Whew. I only ran the tests on my machine before creating this PR, not the build, and it looks like `enzyme` doesn't support the new types in `react@16.8` yet...

Closing this for now.

PR #731, opened 2019-05-12
Update Supercluster dependency as is in MapboxGL

radeno:
Minimal MapboxGL JS peerDependency for for this package is version 0.54.

MapboxGL JS depends on Supercluster 6 https://github.com/mapbox/mapbox-gl-js/blob/v0.54.0/package.json#L37 but in this React wrapper is still version 3, which makes larger bundle.

This PR makes same version lock.


mklopets:
Thanks for pointing this out! This breaks the examples on my machine though. Will look into it tomorrow.

radeno:
@mklopets did you get same error? `supercluster_namespaceObject is not a constructor` ?
Should be Webpack builder in example package. Because in builded files that variable doesn't exists.

alex3165:
This is breaking the example for me too, here is what I have:
Screenshot 2019-10-19 at 16 01 45

I had a quick look and tried different import but still break on example, I am not too sure of why Webpack can't find the module. Also tried with react-scripts since the examples on master use react-scripts now but still not working. I would not merge this PR until we figure out what is going on, otherwise i agree it would be nice to bump supercluster since this library is far behind.

alex3165:
Seems like importing with `import Supercluster from 'supercluster';` fix the examples but break the tests.

radeno:
@alex3165 https://github.com/Wykks/ngx-mapbox-gl/issues/146
Adding `allowSyntheticDefaultImports` should helps. It doesn't change compiling, just type checking.
`@types/supercluster` and `supercluser` package are inconsistent.

radeno:
@alex3165 @mklopets fixed by adding `esModuleInterop` into `tsconfig.json`

alex3165:
Tested and seems to be working, I will get this merged and released, thanks for your work

PR #764, opened 2019-08-16
Rename unsafe lifecycle methods.

johnnynia:
After updating to React v16.9.0 the console is cluttered with warnings about the deprecated use of unsafe lifecycle methods (`componentWillMount`, `componentWillReceiveProps`, `componentWillUpdate`).

The old names will still work in upcoming 16.x and 17.x releases, but in order to let them stand out (and get rid of the console warnings) they have to be prefixed with `UNSAFE_`.

https://reactjs.org/blog/2019/08/08/react-v16.9.0.html#renaming-unsafe-lifecycle-methods

### DIFF
--- a/example/src/demos/allShapes.tsx
+++ b/example/src/demos/allShapes.tsx
@@ -77,7 +77,7 @@ class AllShapes extends React.Component<Props, State> {
   private timeoutHandle: any = undefined;
   private mounted = false;
 
-  public componentWillMount() {
+  public UNSAFE_componentWillMount() {
     this.mounted = true;
     this.timeoutHandle = setTimeout(() => {
       if (this.mounted) {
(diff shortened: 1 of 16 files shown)


vince1995:
Nice, the warnings in the console are really annoying.

muresanandrei:
So any updated on official update with this  I would like to have alex master version to keep future updates in synch.

schrapel:
I think we should ideally solve the root issue here than continue to use these functions.

redbmk:
@schrapel I agree long term, but this is a quick fix and a non-breaking change. I'd imagine `componentDidMount` and `componentDidUpdate` would work, but since they're not quite the same I'm sure there would be some differences.  Hooks might also make this simpler, but that would be a pretty big change, as well as require React v16.8 meaning it should be a major version bump.

Would be great to at least get this in now as a patch release just to avoid the annoying console warnings.

alex3165:
Agree for this as quick fix for now, I will get this merged and released as patch

PR #765, opened 2019-08-19
Fix layer error on style change

aratcliffe:
A "Style is not done loading" error is thrown if a layer's props are updated while a new map style is being loaded. This fix checks to see that the layer is still on the map before working with it in componentWillReceiveProps()


erinmeyers88:
I am having the same issues. Is there a plan to merge this fix? Thanks

Issue #652, opened 2018-10-31
feature.id missing from dataset


daviestar:
Hey,

Trying to implement this example:
https://www.mapbox.com/mapbox-gl-js/example/hover-styles

It uses the map method `setFeatureState`:
https://www.mapbox.com/mapbox-gl-js/api/#map#setfeaturestate

which states:

> This method requires the `feature.id` attribute on data sets.

I can see from the event object that this library is automatically setting a `feature.properties.id`, but unfortunately that doesn't appear compatible with `map.setFeatureState` which needs `feature.id`.

For example this code has no effect:
```js
handleLineMouseEnter = (e) => {
  const {map, feature} = e
  const {properties: {id}, source} = feature
  map.setFeatureState({id, source}, {hover: true})
}

// ...

<Layer
  type="line"
  paint={{
    'line-color': [
      'case', ['boolean', ['feature-state', 'hover'], false],
      '#ff0000',
      '#ff8f33'
    ]
  }}>
  {lines.map((line, idx) => (
    <Feature
      key={idx}
      coordinates={line}
      onMouseEnter={this.handleLineMouseEnter}
    />
  ))}
</Layer>
```

My naeve suggestion would be that if a user explicitly sets an `id` prop on the feature, then it could be added to the dataset, ie:

```js

handleLineMouseEnter = (e) => {
  console.log(e.feature.id) // line-a
}

<Feature
    id="line-a"
    coordinates={line}
    onMouseEnter={this.handleLineMouseEnter}
/>
```

qlerebours:
@daviestar I opened a PR to fix this.
You can patch the plugin while this is not merge. Your code should now work as expected (ofc, I tested it before submitting the PR).

Here is another example that can achieve what you wanted to do (hovering features)

Function called onStyleLoad event:
```
  onMapLoaded = (map) => {
    this.map = map;
    // This is a copy paste of mapbox code
    this.map.on("mousemove", "tm-countries", (e) => {
      if (e.features.length > 0) {
        if (this.hoveredCountry) {
          this.map.setFeatureState({source: 'countries-source', id: this.hoveredCountry}, { hover: false});
        }
        this.hoveredCountry = e.features[0].id;
        this.map.setFeatureState({source: 'countries-source', id: this.hoveredCountry}, { hover: true});
      }
    });

    this.map.on("mouseleave", "tm-countries", () => {
      if (this.hoveredCountry) {
[...]

Issue #603, opened 2018-06-22
GeoJSONLayer layerOptions.id error


dm261395:
`@types/mapbox-gl": "0.45.0"`
`mapbox-gl": "0.46.0"`
`react-mapbox-gl": "3.8.0"`

Hello. I have discovered a bug related to passing a layerOptions prop to the GeoJSONLayer component, specifically when setting the `id` prop of the layerOptions object.

In `geojson-layer.ts`, within the [createLayer](https://github.com/alex3165/react-mapbox-gl/blob/master/src/geojson-layer.ts#L135) function, we see that `map.addLayer` is [called](https://github.com/alex3165/react-mapbox-gl/blob/master/src/geojson-layer.ts#L150) when creating each internal layer for the GeoJSONLayer [(symbol](https://github.com/alex3165/react-mapbox-gl/blob/master/src/geojson-layer.ts#L197) layer, line layer, etc.).

In particular, by [spreading](https://github.com/alex3165/react-mapbox-gl/blob/master/src/geojson-layer.ts#L157) the layerOptions object, the ID for the layer being created is overwritten with the ID of the layerOptions object. This causes the error `error: Error: Layer with id <id> already exists on this map`, since multiple layers are being added to the map with the same ID.

The layerOptions prop is typed as `Layer` from MapboxGL, which requires an `id` prop. This does not seem appropriate.

This also caused my symbol mouse handlers to not work, and this may be the cause of previous issues https://github.com/alex3165/react-mapbox-gl/issues/483.

admataz:
same error here, example: 

```
<GeoJSONLayer
              data={countriesGeojson}
              fillPaint={{
                "fill-color": "purple",
                "fill-outline-color": "yellow"
              }}
              layerOptions={{
                filter:['==','iso_a3', 'AFG'],
                id:'thisisnotworking', 
                type:"fill",
              }}
            />
```

rlueder:
Possibly related to https://github.com/alex3165/react-mapbox-gl/issues/652 ?

qlerebours:
I fixed it in my local env, I'll make a PR soon

qlerebours:
@dm261395 , this is fixed in my PR, if needed you can use [npm patch package](https://www.npmjs.com/package/patch-package) to patch the lib while this is not merged.

I will also comment #652 that will be fixed with this PR to show the working code

AndrewEastwood:
if it's fixed ... then why it's not merged yet ?

[...]

PR #769, opened 2019-08-25
Fix 652 and 603 by adding an id to the geojson feature for Layer component and prevent creating as many layers as there are layer types for geojsonLayer component

qlerebours:
**As mentioned in #652 :**
An id was missing in the geojson features that was preventing to use this kind of feature: https://docs.mapbox.com/mapbox-gl-js/example/hover-styles/
I added the same id to the feature than the one that was added to the feature properties:
```
{
    type: 'Feature',
    geometry: this.geometry(props.coordinates),
    properties: { ...props.properties, id },
    id
}
```
The example code that were not working in the issue is now working fine

**As mentioned in #603 :**
Passing `layerOptions` props with and `id` to the `GeoJSONLayer` component was not working.
It was because in the `initialize` function, a layer was created for each layer type, even if it was stated that it should be one precise type:
```
this.createLayer('symbol');
this.createLayer('line');
this.createLayer('fill');
this.createLayer('fill-extrusion');
this.createLayer('circle');
```
I added a check to create the layer only if the layer type is the right one.
```
if (layerOptions && layerOptions.type && layerOptions.type !== type) {
      return;
}
```
Note that I also documented the `id` props of the `GeoJSONLayer` component that was missing.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -246,6 +246,7 @@ import { GeoJSONLayer } from "react-mapbox-gl";
 ### Properties
 
 * **data** _(required)_: `string | object` The url to the geojson file or the geojson file itself.
+* **id**: `string` An id to give to the source.
 * **lineLayout** | **symbolLayout** | **circleLayout** | **fillLayout** | **fillExtrusionLayout**: `Object` Layer layout information. [mapbox layout api](https://www.mapbox.com/mapbox-gl-style-spec/#layer-layout)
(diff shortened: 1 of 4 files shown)


alex3165:
Hey @qlerebours sorry for the late reply and thank you for your PR, I agree with the ID fix although for the layer type fix I am not sure this is going to work. The problem is that `layerOptions` is an optional parameter, one might use `GeoJsonLayer` and expect his layer to be created. An alternative check could have the following logic:

- If any prop is in `LineProps` (interface) then execute `this.createLayer('line');`
...

I would be happy to merge this PR if you could make this change 👍

solarstar101:
anyone going to bother merging this?

lhammond (approved):
looks good to me.

PR #781, opened 2019-10-01
improve Polygon vs MultiPolygon detection in fill Layer

jbeuckm:
I have MultiPolygons with only one shape inside. These were getting recognized as Polygons and not filling correctly. Here is a fix.


alex3165:
Looking at the [Geojson documentation](http://wiki.geojson.org/GeoJSON_draft_version_6), the correct signature for polygon is `[number, number][][]` where for multipolygon is `[number, number][][][]` in your tests you assume polygon is `[number, number][]` which is not correct. Therefore doing `coordinates.length > 1` is the correct way to detect multi polygon.

alex3165:
Although I really think that unit testing this is very relevant, feel free to open a new PR with the tests if you have some time.

jbeuckm:
> Looking at the [Geojson documentation](http://wiki.geojson.org/GeoJSON_draft_version_6), the correct signature for polygon is `[number, number][][]` where for multipolygon is `[number, number][][][]` in your tests you assume polygon is `[number, number][]` which is not correct. Therefore doing `coordinates.length > 1` is the correct way to detect multi polygon.

Did you read the issue? `coordinates.length > 1` fails the test I provided.

jbeuckm:
> Therefore doing `coordinates.length > 1` is the correct way to detect multi polygon.

What happens if the MultiPolygon has only one Polygon in the array?

alex3165:
Ah yeah that would fail indeed, I think your fix would work if you do `Array,isArray(coordinates[0][0][0])` then it would be a MultiPolygon. Could you re open a PR with this update?

alex3165:
Indeed that would not work in that case, from what I understood your fix is actually valid if you do `Array.isArray(coordinates[0][0][0])` then it would be a multipolygon. Could you update and Reopen a PR?

jbeuckm:
> Indeed that would not work in that case, from what I understood your fix is actually valid if you do `Array.isArray(coordinates[0][0][0])` then it would be a multipolygon. Could you update the PR?

👍🏼Will do. I see what you mean about the nesting levels Polygon vs MultiPolygon - looking for the place where I ran into trouble to verify whether I had the structure wrong on that end.

jbeuckm:
Fixed nesting levels on Polygon vs MultiPolygon

jbeuckm:
This branch is a little stale - started a new PR (that is failing at the moment). https://github.com/alex3165/react-mapbox-gl/pull/790

PR #787, opened 2019-10-19
Fix max-width property name typo in home.tsx

krisaoe:


alex3165:
The title is not centered with `max-width` applied, I will just remove this line, thanks for pointing it out

PR #790, opened 2019-10-21
Improve Polygon vs MultiPolygon Geometry for Layer Fill

jbeuckm:


alex3165:
Thank you for your work, merging this and will release asap

PR #794, opened 2019-10-23
Update Mapbox from v1.1.1 to v1.4.1

kjkurtz:
This PR updates the mapbox-gl dependency from 1.1.1 to 1.4.1.

Mapbox added these features in v1.4.0 which can prove handy especially for labels:

- Add image expression operator to determine image availability (#8684)
- Enable text-offset with variable label placement (#8642)

As well is these in v1.3.0

- Introduce text-writing-mode symbol layer property to allow placing point labels vertically. #8399
- Extend variable text placement to work when text/icon-allow-overlap is set to true. #8620
- Allow text-color to be used in formatted expressions to be able to draw different parts of a label in different colors. #8068

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-mapbox-gl",
-  "version": "4.7.0",
+  "version": "4.7.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


mklopets:
Will update to the latest mapbox-gl (now 1.6.0)

PR #808, opened 2019-12-06
chore: Export props as separate types

olso:
@alex3165 @mklopets 

Having trouble customizing the components without these types

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,17 +1,20 @@
 // Add a style tag to the document's head for the map's styling
-import Map from './map';
-import BaseLayer from './layer';
+import Map, { Props as MapProps } from './map';
+import { Events as MapEvents } from './map-events';
+import BaseLayer, { Props as BaseLayerProps } from './layer';
 import layerMouseTouchEvents from './layer-events-hoc';
-import GeoJSONLayer from './geojson-layer';
-import Feature from './feature';
-import ZoomControl from './zoom-control';
-import Popup from './popup';
-import ScaleControl from './scale-control';
-import Marker from './marker';
-import Source from './source';
-import Cluster from './cluster';
-import RotationControl from './rotation-control';
-import Image from './image';
+import GeoJSONLayer, { Props as GeoJSONLayerProps } from './geojson-layer';
+import Feature, { Props as FeatureProps } from './feature';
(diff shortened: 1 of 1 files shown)


mklopets:
Thanks for this. I'm afk but will release it over the weekend.

olso:
@mklopets nice 😇

mklopets:
Hey, I took another look at this and did a lil' [Googling](https://stackoverflow.com/a/55005902/3790987) and found something I wasn't aware of – there's `React.ComponentProps<typeof ComponentName>` that works for both class as well as function components.

This should let you do whatever you needed these types for originally – do let me know if I'm missing something – but I'm going to close the PR with the assumption that the exports aren't needed and we can keep the codebase a very tiny bit cleaner by omitting these.

olso:
Nice find @mklopets

olso:
@mklopets Unfortunately, doesn't seem to work at all 😞

```tsx
import { Cluster } from "react-mapbox-gl";

type ClusterProps = React.ComponentProps<typeof Cluster>;
```

returns `unknown`

```
npx tsc --version
Version 3.7.3
```

mklopets:
Huh... yep, I can reproduce that – at a quick glance I couldn't really figure out what's causing the issue but I did find a fix:

importing from `react-mapbox-gl/lib/cluster`, `react-mapbox-gl/lib-esm/cluster` and `react-mapbox-gl/src/cluster` all work for getting the right typing through

something's wrong with the entrypoint – if you happen to figure out what it is, I'd be grateful 😄 but hope the workaround works for you.

olso:
@mklopets importing types from `src` was my initial solution, but I couldn't compile because it was full of type errors

`lib-esm` seems to do the trick!

however, it gets messy real fast

```tsx
import ReactMapboxGl from "react-mapbox-gl";
import { Props as MapProps } from "react-mapbox-gl/lib-esm/map";
import { Events as MapEvents } from "react-mapbox-gl/lib-esm/map-events";
```

btw my PR fixes this nicely, no additional fiddling around needed

mklopets:
I'd like to avoid cluttering the exports unnecessarily – the core problem here to be fixed here is the broken entrypoint

PR #820, opened 2020-01-17
Update listeners when event props change

sanfilippopablo:
The code right now checks:
- If there was a listener and got removed, then unlisten from the map.
- If there wasn't a listener before and one was added to the props, add it to the map.
But it's not checking if the listener simply changed. This fixes that issue, making the map be always up to date with the listeners passed through props.


olso:
@alex3165

dimagimburg:
Any updates on this? right now I have to manage complicated code using refs everywhere as a workaround.

mklopets:
Thanks @sanfilippopablo! Released as [v4.8.4](https://github.com/alex3165/react-mapbox-gl/releases/tag/v4.8.4)

PR #910, opened 2020-11-11
Introducing MapboxProvider

jackall3n:
The PR is a proposal for the implementation of the Context API within `react-mapbox-gl`. 

The intention of the the `<MapboxProvider />` is to replace `const Map = ReactMapboxGl({ accessToken: token });` with an implementation like below:

```typescript
import React from 'react';
import ReactDOM from 'react-dom';
import { Map, MapboxProvider } from 'react-mapbox-gl';

ReactDOM.render(
  <MapboxProvider accessToken={process.env.MAPBOX_ACCESS_TOKEN}>
    <Map style="" />
  </MapboxProvider>,
  document.getElementById('root')
)
```

I'm keen to hear the thoughts of the maintainers before I go through and update examples/documentation.


jackall3n:
Hey @alex3165, what's your thoughts on this approach?

PR #914, opened 2020-11-19
feat: handle updates to Source tileJsonSource prop more gracefully

henrinormak:
In mapbox 1.12.0, [a feature](https://github.com/mapbox/mapbox-gl-js/pull/8048) was released which allows changing the URL/tiles props on a vector source without removing it from the map. This is much more performant, especially in cases where the map undergoes several back-to-back changes to its tiles URL (after a few rapid changes, mapbox-gl starts to have issues when constantly removing/re-adding layers/sources).

In this PR, I've:
- Updated dependency to mapbox-gl to be the latest (I had to make some type changes to get the build to pass, ideally these wouldn't be needed, but I'm hoping these could be resolved at a later date)
- Modified the `Source` component to use the new `setUrl` and `setTiles` APIs when dealing with a vector source. I kept the old behaviour as a fallback, as raster tile sources for example didn't get this new feature.

When testing in a local setup with an app that does many `tiles` property changes, this new approach is significantly more performant.

### DIFF
--- a/src/source.ts
+++ b/src/source.ts
@@ -124,7 +124,11 @@ export class Source extends React.Component<Props> {
 
         const hasNewSourceTiles =
           tileJsonSource.tiles !== this.props.tileJsonSource.tiles;
-        if (hasNewSourceTiles && this.props.tileJsonSource.tiles !== undefined) {
+
+        if (
+          hasNewSourceTiles &&
+          this.props.tileJsonSource.tiles !== undefined
+        ) {
           source.setTiles(this.props.tileJsonSource.tiles);
           tilesUpdated = true;
         }
--- a/src/source.ts
+++ b/src/source.ts
@@ -117,14 +117,14 @@ export class Source extends React.Component<Props> {
         const hasNewSourceUrl =
           tileJsonSource.url !== this.props.tileJsonSource.url;
 
-        if (hasNewSourceUrl && this.props.tileJsonSource.url) {
+        if (hasNewSourceUrl && this.props.tileJsonSource.url !== undefined) {
           source.setUrl(this.props.tileJsonSource.url);
           urlUpdated = true;
         }
 
         const hasNewSourceTiles =
           tileJsonSource.tiles !== this.props.tileJsonSource.tiles;
-        if (hasNewSourceTiles && this.props.tileJsonSource.tiles) {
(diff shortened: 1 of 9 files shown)


src/source.ts
```diff
@@ -108,13 +106,37 @@ export class Source extends React.Component<Props> {
...
     // Update tilesJsonSource
     if (tileJsonSource && this.props.tileJsonSource) {
+      let urlUpdated = false;
+      let tilesUpdated = false;
+
+      if (source && source.type === 'vector') {
+        const hasNewSourceUrl =
+          tileJsonSource.url !== this.props.tileJsonSource.url;
+
+        if (hasNewSourceUrl && this.props.tileJsonSource.url) {
```
mklopets, line 120: this doesn't permit unsetting the URL – not sure if there would be a valid use case for it, but is it intentional here?

src/source.ts
```diff
@@ -108,13 +106,37 @@ export class Source extends React.Component<Props> {
...
+          source.setUrl(this.props.tileJsonSource.url);
+          urlUpdated = true;
+        }
+
+        const hasNewSourceTiles =
+          tileJsonSource.tiles !== this.props.tileJsonSource.tiles;
```
mklopets, line 126:
```suggestion
          tileJsonSource.tiles !== this.props.tileJsonSource.tiles;
          
```

for readability

src/source.ts
henrinormak: The Mapbox types seem to indicate that the url cannot be "unset", i.e the functions don't take undefined as an argument - for that I guess one would have to change the entire source. I did now however make this check more explicit about undefined.

henrinormak:
Added an example as well, based it on [the example from Mapbox](https://docs.mapbox.com/mapbox-gl-js/example/third-party/), with the extension that you can switch the mirror for the tiles, which also tests the changes in this PR.

The problem I had originally is not surfaced in this demo, unfortunately, as likely the variety in the tiles URL is too small (i.e if it had query parameters that changed etc, then the problem would surface), I did manage to reproduce it by adding a random query parameter to it (something like `t=Math.random()`), after toggling the layer a few times, the performance of reloading the tiles after the button clicked started going down - this did not happen with the updated `Source`

mklopets (approved):

PR #926, opened 2020-12-31
Defines idle event handler and corresponding property callback

### LINKED ISSUE #925: map.on('idle') event doesn't seem to be supported as onIdle
I don't see support for the `idle` map event e.g. I can't do:
```
<ReactMapboxGL onIdle={myIdleHandler}/>
```

Idle does seem like a valid MapBoxGL event: https://docs.mapbox.com/mapbox-gl-js/api/map/#map.event:idle

There's no `onIdle` MapEvent defined here: 
https://github.com/alex3165/react-mapbox-gl/blob/6507a2273a6680ae95b5b1379db8f41abc379f61/src/map-events.tsx#L9-L53

snickell:
Fixes https://github.com/alex3165/react-mapbox-gl/issues/925

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -128,7 +128,8 @@ const events = {
   onBoxZoomCancel: 'boxzoomcancel',
   onRotateStart: 'rotatestart',
   onRotate: 'rotate',
-  onRotateEnd: 'rotateend'
+  onRotateEnd: 'rotateend',
+  onIdle: 'idle',
 };
 ```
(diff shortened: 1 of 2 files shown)


thomashermine (commented):
I tested this PR locally, and it works like a charm. 
It does seems like a simple omission, any chance this could be merged? 😊

massimopalmieri (approved):

PR #947, opened 2021-04-22
updated react peer dep versions/mapbox versions

danditomaso:


danditomaso:
No response from author. Closing PR

Issue #927, opened 2021-01-04
Map callbacks not working with useState hook after version 4.8.6


wtanna:
It seems that version 5 has broken the callbacks for the `<Map />` component when using the `useState` hook. Here is a simplified example map that uses `onMoveStart`, `onMoveEnd` with a `Marker` that also handles the same piece of state. The below code can be put into a https://codesandbox.io/ after adding the following packages:

- mapbox-gl  @2.0.1
- react-mapbox-gl @5.1.1

```jsx
import React, { useState } from "react";
import "./styles.css";

import ReactMapboxGl, { Marker } from "react-mapbox-gl";
import "mapbox-gl/dist/mapbox-gl.css";

const appStyles = {
  display: "flex",
  "flex-direction": "column",
  "text-align": "center",
  "font-family": "sans-serif",
  height: "100vh",
  width: "100vw"
};

const markerStyles = {
  "background-color": "pink",
  display: "flex",
  "align-items": "center",
  height: "50px",
  "font-size": "20px"
};

const mapContainerStyles = {
  flex: "1 1 auto"
};

// Make sure you put in your access token
const Map = ReactMapboxGl({
  accessToken: ACCESS_TOKEN
  dragRotate: false,
  pitchWithRotate: false
});

export default function App() {
  const [count, setCount] = useState(0);
  const [stylesLoaded, setStylesLoaded] = useState(false);

  return (
    <div style={appStyles} className="App">
      <h1>Count: {count}</h1>
      <h3>Styles Loaded: {JSON.stringify(stylesLoaded)}</h3>
      <Map
        onMoveStart={() => {
          console.log("onMoveStart count", count);
          setCount(1);
        }}
        onMoveEnd={() => {
          console.log("onMoveEnd count", count);
          if (count > 0) setCount(0);
        }}
        onStyleLoad={() => {
          setStylesLoaded(true);
        }}
        containerStyle={mapContainerStyles}
        style="mapbox://styles/mapbox/streets-v11"
      >
        <Marker
          style={markerStyles}
          coordinates={[-0.2416815, 51.5285582]}
          anchor="bottom"
          onClick={() => {
            console.log("Marker count", count);
            setCount(2);
          }}
        >
          <div className="marker-content">A Marker</div>
        </Marker>
      </Map>
    </div>
  );
}
```

[...]

PR #948, opened 2021-04-28
Fix memoized rebind

akre54:
This is a strawman fix for #927. Currently react-mapbox-gl binds DOM events the first time a prop function is passed in, which means that for memoized functions (for instance, with useMemo / React functional components) the handler may be a stale closure and have out-of-date data.

The issue is that the `updateEvents` function in map-events.tsx only checks for the presence of the function but not referential equality, and doesn't dereference the latest value off of `this.props`.

As I see it, there are 3 approaches we could take to fix this in library code:

1. Unbind stale functions and re-bind new functions when passed in (i.e. change the `toListenOff` expression to check against prevProps -- see the first commit here).
2. 3. Dereference the latest value of the callback in this.props (the approach I took here in the second commit). The issue with this approach is that we need access to the `this` object in `updateEvents`, because the `props` changes between renders, so this is a bit uglier than I'd like.
3. Use a Ref, always set to the latest passed-in prop value. This seems like a duplication of 2 with extra steps.

The last approach would be to migrate map.tsx to be a functional component, but that might be a bigger ask. 

Any suggestions here? Thanks!

### DIFF
--- a/src/map-events.tsx
+++ b/src/map-events.tsx
@@ -105,7 +105,7 @@ export type Listeners = {
 
 export const listenEvents = (
   partialEvents: EventMapping,
-  props: Partial<Events>,
+  props: Partial<Events> & { memoizedEvents?: boolean },
   map: MapboxGl.Map
 ) =>
   Object.keys(partialEvents).reduce(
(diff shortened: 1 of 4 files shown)


src/map-events.tsx
```diff
@@ -105,17 +106,14 @@ export type Listeners = {
 
 export const listenEvents = (
   partialEvents: EventMapping,
-  props: Partial<Events>,
+  parent: React.Component,
```
akre54, line 109: We can't pass in `props` because then the closure traps an out-of-date value when the component re-renders. It needs to always grab the latest value off of the `this` object.

PR #1003, opened 2022-05-19
fix: add missing children prop to Popup

Boren:
React 18 requires all components to explicitly specify children prop


gregegan:
Why was this closed? seems valid

